PR #3, opened 2025-10-08
chore: Add presubmit workflows

brianquinlan:
Adds the following workflows:
1. dart analyze
2. dart format (fail on any formatting differences)
3. dart test
4. verify that any generated code matches what would be generated in main

Also ran "dart format ." so that the "dart format" check passes.


devoncarew (commented):
Can you add a PR description and update the action references to their corresponding SHAs? Thanks!

devoncarew (approved):

PR #12, opened 2025-10-10
fix(google_cloud_ai_generativelanguage_v1): Fix tips in README.md

brianquinlan:
The only non-generated change is to .sidekick.toml


devoncarew (approved):
👍

PR #22, opened 2025-10-14
feat: generate three firebase discovery doc based APIs

devoncarew:
Generate three firebase discovery doc based APIs:

- add definitions for fcm.v1.json (Firebase Cloud Messaging), firebaseappcheck.v1.json (firebase AppCheck), and firebaserules.v1.json (firebase rules)
- regenerate the packages

This is to exercise discovery doc support in sidekick. Note that I have local changes to sidekick (a bug fix, updates to the dart generation templates, ...) to generate what's in this PR. We'll need to upstream fixes and bug reports to sidekick before something like this can be checked in.


devoncarew:
I haven't tried using these APIs, but I'm now able to generate all three. The patches to sidekick may not all be landable as-is, but I'll put them somewhere (other than my local disk) for visibility.

[sidekick.patch](https://github.com/user-attachments/files/22916793/sidekick.patch)

PR #32, opened 2025-10-15
docs(google_cloud_ai...): Add a note about model names

### LINKED ISSUE #16: [google_cloud_ai_generativelanguage_v1] add a likely user footgun to the readme
For package:google_cloud_ai_generativelanguage_v1, when porting code from the previous package:google_generative_ai, I was receiving http 404 errors for all my requests. It took a while to track down; I was passing in `gemini-2.5-flash` for the model name; the correct value is `models/gemini-2.5-flash`. The previous version of this client would auto-promote the model names to be prefixed w/ `'model/'` if necessary.

I don't think we can modify the generated code to do checks here, and the failure mode is pretty unhelpful (the model name goes in the url, so we were really seeing 404s back).

We should probably but in a note in the readme for this package, or a general FAQ, or a brief section about migrating from the previous package.

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/16


devoncarew (approved):
👍

Issue #17, opened 2025-10-13
generativelanguage/v1beta does not generate entire service


brianquinlan:
At least the following items are missing:

1. [`thinking_config`](https://github.com/googleapis/googleapis/blob/6821943108fe3284f483defc9b60774a3752de2b/google/ai/generativelanguage/v1beta/generative_service.proto#L523) field and the corresponding message [`ThinkingConfig`](https://github.com/googleapis/googleapis/blob/6821943108fe3284f483defc9b60774a3752de2b/google/ai/generativelanguage/v1beta/generative_service.proto#L297)
2. [`media_resolution`](https://github.com/googleapis/googleapis/blob/6821943108fe3284f483defc9b60774a3752de2b/google/ai/generativelanguage/v1beta/generative_service.proto#L527C28-L527C44) field and the corresponding enum [`MediaResolution`](https://github.com/googleapis/googleapis/blob/6821943108fe3284f483defc9b60774a3752de2b/google/ai/generativelanguage/v1beta/generative_service.proto#L325)

PR #39, opened 2025-10-16
chore: Update to the latest API sources

brianquinlan:
Also adds instructions in DEVELOPER_GUIDE.md on how to do that (this is the only manual change).

Fixes https://github.com/googleapis/google-cloud-dart/issues/17

### DIFF
--- a/generated/google_cloud_longrunning/lib/longrunning.dart
+++ b/generated/google_cloud_longrunning/lib/longrunning.dart
@@ -96,6 +96,8 @@ final class Operations {
       if (request.filter != null) 'filter': request.filter!,
       if (request.pageSize != null) 'pageSize': '${request.pageSize}',
       if (request.pageToken != null) 'pageToken': request.pageToken!,
+      if (request.returnPartialSuccess != null)
+        'returnPartialSuccess': '${request.returnPartialSuccess}',
     });
     final response = await _client.get(url);
     return ListOperationsResponse.fromJson(response);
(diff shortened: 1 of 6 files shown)


devoncarew:
As a drive by comment, it's interesting how much new API reving `googleapis/googleapis` can bring in.

generated/google_cloud_longrunning/lib/longrunning.dart
```diff
@@ -96,6 +96,8 @@ final class Operations {
       if (request.filter != null) 'filter': request.filter!,
       if (request.pageSize != null) 'pageSize': '${request.pageSize}',
       if (request.pageToken != null) 'pageToken': request.pageToken!,
```
natebosch: [out of scope for this PR] Is this generated or hand-written? Consider a followup refactoring to pattern conditionals to avoid the `!` operators.

```dart
      if (request.filter case final filter?) 'filter': filter,
      if (request.pageSize != null) 'pageSize': '${request.pageSize}',
      if (request.pageToken case final pageToken?) 'pageToken': pageToken,
```

generated/google_cloud_longrunning/lib/longrunning.dart
```diff
@@ -201,11 +203,24 @@ final class ListOperationsRequest extends ProtoMessage {
   /// The standard list page token.
   final String? pageToken;
 
+  /// When set to `true`, operations that are reachable are returned as normal,
```
natebosch: [nit] Consider rephrasing to start with "Whether".  https://dart.dev/effective-dart/documentation#prefer-starting-a-boolean-variable-or-property-comment-with-whether-followed-by-a-noun-or-gerund-phrase

PR #46, opened 2025-10-17
chore(gax): Use repository analysis_options

### LINKED ISSUE #21: Remove analysis options from all subpackages so only the root one is used
I missed the one in gax for sure but there might be others.

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/21


natebosch (approved):
I'm skeptical of using the lint that requires `const`. If this code runs on the web there may be reasons to avoid making everything const eagerly.

I don't think this needs to blocked for now, but we might want to look in to why that lint is enabled here.

brianquinlan:
> I'm skeptical of using the lint that requires `const`. If this code runs on the web there may be reasons to avoid making everything const eagerly.
>
> I don't think this needs to blocked for now, but we might want to look in to why that lint is enabled here.

I just copied these lints from `package:http`. Kevin enabled them a long time ago:
https://github.com/dart-lang/http/commit/a8efbef05a58919dc7aa2dab42198334f2459ffb

PR #48, opened 2025-10-21
fix: Remove gax dependency on `package:google_cloud_rpc`

brianquinlan:


brianquinlan:
Abandoned in favor of https://github.com/googleapis/google-cloud-dart/pull/50

PR #50, opened 2025-10-21
fix: Remove the `_gax` dependency on `_rpc`

brianquinlan:
Moves:
- `Status` from `_rpc` to `_gax` (in `status.dart`)
- `Any` from `protobuf` to `_gax` (in `any.dart`)
- Moved `JsonEncodable`/`ProtoMessage`/`ProtoEnum` interfaces into `proto.dart`
- Fixed any lint errors
- exports those symbols from the packages that they were removed from (changes to .sidekick.toml)
- moves relevant tests for `Any` to `_gax`

### DIFF
--- a/packages/google_cloud_gax/lib/src/status.dart
+++ b/packages/google_cloud_gax/lib/src/status.dart
@@ -16,6 +16,10 @@ import 'any.dart';
 import 'encoding.dart';
 import 'proto.dart';
 
+// This class logically belongs in `package:google_cloud_rpc` but is here
+// because it is used by the error handling in `ServiceClient` and we don't want
+// to have circular dependencies between `package:google_cloud_gax` and
+// `package:google_cloud_rpc`.
 /// The `Status` type defines a logical error model that is suitable for
 /// different programming environments, including REST APIs and RPC APIs. It is
 /// used by [gRPC](https://github.com/grpc). Each `Status` message contains
@@ -23,10 +27,6 @@ import 'proto.dart';
 ///
 /// You can find out more about this error model and how to work with it in the
 /// [API Design Guide](https://cloud.google.com/apis/design/errors).
-// This class logically belongs in `package:google_cloud_rpc` but is here
-// because it is used by the error handling in `ServiceClient` and we don't want
-// to have circular dependencies between `package:google_cloud_gax` and
-// `package:google_cloud_rpc`.
 final class Status extends ProtoMessage {
   static const String fullyQualifiedName = 'google.rpc.Status';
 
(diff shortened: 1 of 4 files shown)


generated/google_cloud_protobuf/lib/protobuf.dart
```diff
@@ -22,6 +22,8 @@ library;
 import 'package:google_cloud_gax/gax.dart';
 import 'package:google_cloud_gax/src/encoding.dart';
 
+export 'package:google_cloud_gax/gax.dart' show Any;
```
devoncarew: As this file is generated, you'll need to introduce a way to define exports in a sidekick file.

packages/google_cloud_gax/lib/gax.dart
```diff
@@ -14,59 +14,25 @@
 
 import 'dart:convert';
 
-import 'package:google_cloud_rpc/rpc.dart';
 import 'package:http/http.dart' as http;
-export 'src/web.dart'
-    if (dart.library.io) 'src/vm.dart'
-    show environmentVariable;
 
+import 'src/proto.dart';
```
devoncarew: I like having a `proto.dart` file. Perhaps put 'Any' there, and 'Status' in a corresponding `rpc.dart` file? We could then document where they were from originally.

JsonEncodable, ProtoMessage, and ProtoEnum could then go in some other support file?

brianquinlan: I created:
- `proto.dart`
- `any.dart`
- `status.dart`

packages/google_cloud_gax/pubspec.yaml
```diff
@@ -27,9 +27,10 @@ environment:
 resolution: workspace
 
 dependencies:
-  google_cloud_rpc: ^0.1.0
   http: ^1.3.0
 
 dev_dependencies:
+  google_cloud_protobuf: ^0.1.0
```
devoncarew: Perhaps add a comment that these deps are for tests.

packages/google_cloud_gax/lib/src/status.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'proto.dart';
+
+/// The `Status` type defines a logical error model that is suitable for
+/// different programming environments, including REST APIs and RPC APIs. It is
+/// used by [gRPC](https://github.com/grpc). Each `Status` message contains
+/// three pieces of data: error code, error message, and error details.
+///
+/// You can find out more about this error model and how to work with it in the
+/// [API Design Guide](https://cloud.google.com/apis/design/errors).
+// This class logically belongs in `package:google_cloud_rpc` but is here
```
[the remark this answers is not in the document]
brianquinlan: Done.

PR #58, opened 2025-10-24
feat: use proto presence information to inform nullability

devoncarew:
An update to https://github.com/googleapis/google-cloud-dart/pull/43; this is still a draft PR.

- now use whether a field is in a oneof to help inform nullability

### DIFF
--- a/generated/google_cloud_ai_generativelanguage_v1beta/lib/generativelanguage.dart
+++ b/generated/google_cloud_ai_generativelanguage_v1beta/lib/generativelanguage.dart
@@ -1767,7 +1767,7 @@ final class GetCachedContentRequest extends ProtoMessage {
 
   @override
   Object toJson() {
-    return {if (name.isNotDefault) 'name': name};
+    return {'name': name};
   }
 
   @override
@@ -1827,7 +1827,7 @@ final class DeleteCachedContentRequest extends ProtoMessage {
 
   @override
   Object toJson() {
-    return {if (name.isNotDefault) 'name': name};
+    return {'name': name};
   }
 
   @override
@@ -2389,7 +2389,7 @@ final class FileData extends ProtoMessage {
   Object toJson() {
     return {
       if (mimeType.isNotDefault) 'mimeType': mimeType,
-      if (fileUri.isNotDefault) 'fileUri': fileUri,
+      'fileUri': fileUri,
     };
   }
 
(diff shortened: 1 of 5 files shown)


generated/google_cloud_ai_generativelanguage_v1beta/example/main.dart
```diff
@@ -27,11 +27,11 @@ void main() async {
   );
 
   final result = await service.generateContent(request);
-  final parts = result.candidates?[0].content?.parts;
+  final parts = result.candidates.first.content?.parts;
   if (parts == null) {
     print('<No textual response>');
   } else {
-    print(parts.map((p) => p.text ?? '').join(''));
+    print(parts.map((p) => p.text).join(''));
```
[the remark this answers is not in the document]
devoncarew: No - good catch. `text` is nullable, but we're assuming has content based on the context of the sample. I added a 'bang' (`!`).

```dart
print(parts.map((p) => p.text!).join(''));
```

generated/google_cloud_ai_generativelanguage_v1beta/example/application_default_credentials.dart
```diff
@@ -45,11 +45,11 @@ void main() async {
   );
 
   final result = await service.generateContent(request);
-  final parts = result.candidates?[0].content?.parts;
+  final parts = result.candidates.first.content?.parts;
```
brianquinlan:
```suggestion
  final parts = result.candidates.firstOrNull?.content?.parts;
```

generated/google_cloud_ai_generativelanguage_v1beta/example/streamed_response.dart
```diff
@@ -29,9 +29,9 @@ void main() async {
   );
 
   await for (final result in service.streamGenerateContent(request)) {
-    final parts = result.candidates?[0].content?.parts;
+    final parts = result.candidates.first.content?.parts;
     if (parts != null) {
-      parts.forEach((p) => stdout.write(p.text ?? ''));
+      parts.forEach((p) => stdout.write(p.text));
```
brianquinlan:
```suggestion
      parts.forEach((p) => stdout.write(p.text!));
```

I think? Could you rerun these examples.

brianquinlan (approved):
This is going to be a pain to merge.

For package versions, I think that the easiest thing to do is:
1. remove all of the `version = ` from all .sidekick.toml except the one at the root
2. update the root .sidekick.toml "# Generated packages section" such that all of the version constraints are '^0.2.0'

Not only is that expedient, but I think it is also correct ;-)

Let me know if you want me to do the merge work.

PR #67, opened 2025-10-30
feat: Move `_gax` functionality to `_rpc`

brianquinlan:
Changes:
- move `_gax` functionality and tests to `_rpc`
- deprecate `_gax`
- update various .sidekick.toml files
- remove `_gax` tests from CI

Relies on https://github.com/googleapis/librarian/pull/2713


generated/google_cloud_rpc/lib/service_client.dart
```diff
@@ -0,0 +1,249 @@
...
+export 'dart:typed_data' show Uint8List;
+
+export 'src/web.dart'
+    if (dart.library.io) 'src/vm.dart'
+    show httpClientFromApiKey;
+
+const String _clientKey = 'x-goog-api-client';
+
+// ignore: prefer_const_declarations
+final String _clientName = 'gl-dart/$clientDartVersion gax/$gaxVersion';
```
devoncarew: Just double checking that we still want our client name to contain 'gax' (`gl-dart/0.x.0 gax/0.2.0`).

devoncarew (approved):
lgtm

I like the gax deprecation + re-export of rpc technique. One question in-line about the client name.

We now only have packages in `generated/`. Longer term, we could keep packages there, move them to `packages/`, or move them to something like `pkgs/`. The split had originally been to keep generated packages separate from hand-written ones for hygiene. Not sure if we'll end up w/ any other hand-written ones living in this repo (auth? firebase?).

natebosch (approved):

PR #68, opened 2025-10-30
feat: add isNotDefault extension methods for core proto types

devoncarew:
- add `isNotDefault` extension methods for core proto types

This is the hand-written portions of https://github.com/googleapis/google-cloud-dart/pull/58. I'll follow this up w/ generator changes (in the librarian repo), and then a PR back to this repo that will re-generate the packages and update test and example code (the rest of #58).

Happy to bike-shed on the extension method name(s). It's `isNotDefault` now, but could be something like `notDefault`, ... .

cc @natebosch


generated/google_cloud_protobuf/lib/src/encoding.dart
```diff
@@ -155,3 +155,39 @@ Map<T, Object?>? encodeMap<T>(Map<T, JsonEncodable>? value) =>
 /// Encode a list of `bytes` values into JSON.
 Map<T, String>? encodeMapBytes<T>(Map<T, Uint8List>? value) =>
     value?.map((key, value) => MapEntry(key, base64Encode(value)));
+
+/// Extensions methods used for comparing to proto default values.
+extension BoolExtension on bool {
```
brianquinlan: Maybe these should be something like `*ProtoDefault` since they will appear in generated documentation.

devoncarew (commented):
> I think that you need to add a version to the .sidekick.toml file for this package.

I incremented the protobuf package version from 0.1.0 to 0.1.1 (and updated the dep other packages have on it in the main sidekick file).

PR #69, opened 2025-10-31
feat: update using null-safety generator change

devoncarew:
Update the packages here using the null-safety generator change:

- a follow up to https://github.com/googleapis/librarian/pull/2726
- update sidekick version reference
- re-generate the code
- update the hand-written code (part files, examples, and tests)

Previously as a draft at https://github.com/googleapis/google-cloud-dart/pull/58; done as separate commits to make review a little easier.

@brianquinlan, this does not include sidekick / dart package version changes. I assume we'll want to bump everything (to 0.2.0?). Not sure if we're bumping on changes or on publish (or don't have a specific process yet).

### DIFF
--- a/generated/google_cloud_protobuf/.sidekick.toml
+++ b/generated/google_cloud_protobuf/.sidekick.toml
@@ -34,7 +34,6 @@ include-list = """\
 """
 
 [codec]
-version          = "0.1.1"
 copyright-year   = '2025'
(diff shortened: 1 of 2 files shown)


devoncarew:
> Could you please remove the version declarations in all of the .sidekick.toml files except the one in the root? And then change the "#Generated packages" section to all be 0.2.0?

Done!

PR #75, opened 2025-11-13
chore: Use sidekick @v1.0.1

brianquinlan:


generated/google_cloud_ai_generativelanguage_v1beta/lib/generativelanguage.dart
```diff
@@ -120,8 +120,8 @@ final class CacheService {
     UpdateCachedContentRequest request,
   ) async {
     final url = Uri.https(_host, '/v1beta/${request.cachedContent!.name}', {
-      if (request.updateMask!.paths.isNotDefault)
-        'updateMask.paths': request.updateMask!.paths,
+      if (request.updateMask != null)
+        'updateMask': request.updateMask!.toJson(),
```
natebosch: Hmm, I missed this in the other PR but there's another option for this output that's more idiomatic in modern Dart.


```suggestion
      if (request.updateMask case final updateMask?)
        'updateMask': updateMask.toJson(),
```

I don't know if it matters enough to warrant another round of changes in the generator but I thought I'd mention it since I always like avoiding `!`

brianquinlan: I do have an issue to fix this but it might not be worth it.

natebosch: The docs for `when` are lower on that page.

https://dart.dev/language/branches#guard-clause

In general our docs for patterns are a little spread out and I always have to dig a bit when I want to finda  specific reference.

PR #84, opened 2025-11-19
feat: Add `google_cloud_logging_*`

brianquinlan:
The hand-written files are:
- //.sidekick.toml
- pubspec.yaml
- .github/workflows/dart_checks.yaml
- generated/google_cloud_logging_v2/test/write_log_test.dart


generated/google_cloud_logging_type/.sidekick.toml
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[general]
+specification-source = 'google/logging/type'
+
+[source]
+title-override = "Logging types"
```
natebosch: [nit] I can't tell if there is a reason behind which strings are single vs double quoted - it doesn't look like any of the double quoted strings are using escape characters. Consider using `'` consistently.

natebosch (approved):

Issue #90, opened 2025-11-26
`decodeList` doesn't work for `List<float>` that happens to contain integral floats


brianquinlan:
See https://github.com/google-gemini/deprecated-generative-ai-dart/issues/234#issuecomment-3578011893

PR #91, opened 2025-11-27
fix: Fix `repeated double` decoding

brianquinlan:
Fixes:
- https://github.com/googleapis/google-cloud-dart/issues/90
- https://github.com/googleapis/google-cloud-dart/issues/92

Changes:
- Upgrade to latest Sidekick (`dart_checks.yaml`)
- Add a test to `...generative_language_...` to check embeddings (`embeddings_test.dart`)
- Change `Operation.fromJson` to take a `Object?` parameter (`longrunning.p.dart`)
- Update encodings to support `uint64` types, remove no-longer-used decoders (and their tests) (`encoding.dart`/`encoding_test.dart`)
- Change "special" decoders (e.g. `Any`) to take `Object?` in their decoders (`protobuf.p.dart`)
- Add some double compliance tests and support "Extreme values" test (`compliance_test.dart`)
- Change test recording to encode the response body as utf8 (rather than base64)
- Regenerate all test recordings
- Update package versions (`.sidekick.toml`)


generated/google_cloud_protobuf/test/encoding_test.dart
```diff
@@ -72,113 +41,6 @@ void main() {
     expect(encodeDouble(double.negativeInfinity), '-Infinity');
   });
 
-  test('enum', () {
```
natebosch: Why are we removing these tests? We should note the removed tests in the commit message.

natebosch (approved):

generated/google_cloud_protobuf/test/encoding_test.dart
brianquinlan: Because I deleted the functions that they are testing. The generator is now doing:

```dart
  field: [for (final i in json['enum']) Enum.FromJson(i)]
```

instead of:

```dart
  field: decodeListEnum(json['enum'], Enum.FromJson)
```

PR #93, opened 2025-12-03
test: Add some conformance tests

brianquinlan:
- Also upgrades to the latest sidekick version (required for `ProtoMessage.toJson()` equality)


generated/google_cloud_showcase_v1beta1/test/showcase_server.dart
```diff
@@ -0,0 +1,123 @@
...
+  @override
+  String toString() => 'PortInUseException: $message';
+}
+
+class ShowcaseServer {
+  final Process _process;
+
+  static Future<void> _install() async {
+    // Install showcase rather than running it using `go run` because `go run`
+    // will then spawn showcase as a subprocess, which means that we don't be
```
natebosch: [nit] 
```suggestion
    // will then spawn showcase as a subprocess, which means that we won't be
```

PR #102, opened 2026-01-06
fix(dart): Add proto serialization conformance tests

brianquinlan:
It adds a not-for-publication `google_cloud_protojson_conformance` package that adds serialization tests based on [this test file](https://github.com/protocolbuffers/protobuf/blob/9eb9b36e8a6f0f2766e0fbb9263035642c66d49e/src/google/protobuf/test_messages_proto3.proto#L41). The tests themselves are in `serialization_test.dart`. I added tests for every field in the test file, though two tests are currently set to skip because they don't pass.

I also added another test to `write_log_test.dart` to test a real use of non-string `map` keys.

`generated/google_cloud_longrunning/.sidekick.toml` had to be modified to explicitly include a dependency on `package:google_cloud_rpc/rpc.dart` because of this fix: https://github.com/googleapis/librarian/pull/3465

Many of the encoders in `generated/google_cloud_protobuf/lib/src/encoding.dart` are no longer used and were removed. But new encoders were added for non-string `map` key types.

`generated/google_cloud_protobuf/lib/src/encoding.dart` remove the tests for the removed encoders, added tests for the new encoders and added additional tests for existing encoders.

### DIFF
--- a/generated/google_cloud_protojson_conformance/.sidekick.toml
+++ b/generated/google_cloud_protojson_conformance/.sidekick.toml
@@ -21,6 +21,6 @@ include-list = 'test_messages_proto3.proto'
 
 [codec]
 dev-dependencies      = 'path,test,test_utils'
-not-for-publication   = 'true'
 library-path-override = 'google_cloud_protobuf_test_messages_proto3.dart'
+not-for-publication   = 'true'
 repository-url        = 'https://github.com/googleapis/google-cloud-dart/tree/main/generated/google_cloud_protojson_conformance'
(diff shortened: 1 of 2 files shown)


generated/google_cloud_protojson_conformance/README.md
```diff
@@ -0,0 +1,25 @@
+## 
+
+The Google Cloud client library for the .
```
natebosch: Should we delete this file or replace it?

brianquinlan: This is generated so it would just get rewritten. I didn't think that is was worth modifying the generator to fix this since it is just a test package.

PR #111, opened 2026-01-28
feat(google_cloud_storage): Bucket creation

brianquinlan:
A follow-up PR will handle the failure case.


generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -344,6 +344,19 @@ extension TimestampExtension on Timestamp {
   }
 }
 
+extension DateTimeToTimestampExtension on DateTime {
+  /// This UTC `DateTime` as a [Timestamp].
+  Timestamp toTimestamp() {
+    if (!isUtc) {
+      throw ArgumentError('DateTime must be in UTC');
```
natebosch: Would StateError make sense here? My understanding is that ArgumentError is only used for parameters, here we are checking the field of what I think we'd call the receiver.

natebosch (approved):

PR #116, opened 2026-01-30
feat(cloud): Add support for updating a bucket

brianquinlan:


packages/google_cloud_storage/lib/src/retry.dart
```diff
@@ -132,6 +136,9 @@ final class ExponentialRetry implements RetryRunner {
       try {
         return await body();
       } catch (e) {
+        if (!isIdempotent) {
+          rethrow;
+        }
```
natebosch: [optional] Consider omitting the braces for short conditionals that can fit on one line. I especially like that style for guard clauses.


```suggestion
        if (!isIdempotent) rethrow;
```

PR #117, opened 2026-02-03
refactor: use `ServiceClient` for transport

brianquinlan:
This PR is in preparation for fixing `Storage.patchBucket`. 

`Storage.patchBucket` cannot be expressed correctly using the storage implementation in `package:googleapis`. Since using `package:googleapis` for API transport without using its JSON serialization, etc., would be difficult, this PR removes the dependency on `package:googleapis` completely.

Other changes:
1. Added JSON serialization logic
2. More extensive tests added to `create_bucket_test.dart` to verify that JSON serialization is correct.

The analyzer failures are for generated code (I need to fix that).

### DIFF
--- a/packages/google_cloud_storage/lib/src/common_json.dart
+++ b/packages/google_cloud_storage/lib/src/common_json.dart
@@ -55,9 +55,7 @@ ObjectAccessControl? objectAccessControlFromJson(Map<String, Object?>? json) {
   );
 }
 
-Map<String, Object?>? objectAccessControlToJson(
-  ObjectAccessControl? instance,
-) {
+Map<String, Object?>? objectAccessControlToJson(ObjectAccessControl? instance) {
   if (instance == null) return null;
   return {
     'bucket': ?instance.bucket,


packages/google_cloud_storage/lib/src/bucket_metadata_json.dart
```diff
@@ -0,0 +1,709 @@
...
+  return BucketIpFilter(
+    allowAllServiceAgentAccess: json['allowAllServiceAgentAccess'] as bool?,
+    allowCrossOrgVpcs: json['allowCrossOrgVpcs'] as bool?,
+    mode: json['mode'] as String?,
+    publicNetworkSource: bucketPublicNetworkSourceFromJson(
+      json['publicNetworkSource'] as Map<String, Object?>?,
+    ),
+    vpcNetworkSources: (json['vpcNetworkSources'] as List<Object?>?)
+        ?.map(
+          (e) => bucketPublicNetworkSourceFromJson(e as Map<String, Object?>?)!,
```
natebosch: This looks a little confusing to cast to a nullable map, then null assert the overall result. It would be more consistent to cast to non-nullable map which is closer to the condition we are expecting, but I can see how that's also a redundant check because we can't remove the `!` anyway.

This line might look more sensible with the extension method refactor I proposed above. As an extension method this looks like `(e as Map<String, Object?>).parseBucketPublicNetworkSourceFromJson()` which I think is an improvement.

PR #120, opened 2026-02-04
chore: regenerate Dart libraries using librarian

JoeWang1127:
Regenerate Dart libraries using librarian.

Fixes https://github.com/googleapis/librarian/issues/3578

### DIFF
--- a/.github/workflows/dart_checks.yaml
+++ b/.github/workflows/dart_checks.yaml
@@ -94,7 +94,7 @@ jobs:
       # the inputs should be pinned, including the generator version and the
       # googleapis SHA.
       - name: Regenerate existing clients
-        # The current head version of sidekick can be generated with:
+        # The current head version of librarian can be generated with:
         # GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main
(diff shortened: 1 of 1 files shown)


librarian.yaml
```diff
@@ -0,0 +1,475 @@
...
+      proto:google.rpc: package:google_cloud_rpc/rpc.dart
+      proto:google.type: package:google_cloud_type/type.dart
+    version: 0.5.0
+libraries:
+  - name: google_cloud_ai_generativelanguage_v1beta
+    apis:
+      - path: google/ai/generativelanguage/v1beta
+    copyright_year: "2025"
+    keep:
+      - example/application_default_credentials.dart
```
brianquinlan: Is there a way to use wildcards here? Maybe applied to the entire repo?

Is there a way to break this into multiple files (like an import statement)? It would be nice to move the readme content to another file.

JoeWang1127:
> Could you update:
https://github.com/googleapis/google-cloud-dart/blob/main/DEVELOPER_GUIDE.md

Can I update the docs in a separate PR, we have an issue for this: https://github.com/googleapis/librarian/issues/3579

PR #121, opened 2026-02-04
fix(gcs): Use a builder for `Storage.patchBucket`

brianquinlan:


packages/google_cloud_storage/lib/src/bucket_metadata_patch_builder.dart
```diff
@@ -0,0 +1,120 @@
...
+
+  BucketMetadataPatchBuilderJsonEncodable(this._builder);
+
+  @override
+  Object? toJson() => _builder._json;
+}
+
+/// A set of fields to update on a [Cloud Storage bucket].
+///
+/// Fields set to `null` are cleared.
```
natebosch: [nit] To avoid confusion over "are fields set to null by default?"

```suggestion
/// Fields explicitly set to `null` are cleared.
```

PR #129, opened 2026-02-06
chore(gcs): Move common test code into a library

brianquinlan:


packages/google_cloud_storage/test/test_utils.dart
```diff
@@ -0,0 +1,32 @@
...
+  for (int i = 0; i < 60; i++)
+    _bucketChars[_random.nextInt(_bucketChars.length)],
+].join();
+
+String bucketName(String name) {
+  assert(name.length < 55, '"$name" is too long to append a random suffix.');
+  return TestHttpClient.isRecording || TestHttpClient.isReplaying
+      ? name
+      : '$name-${uniqueBucketName()}'.substring(0, 60);
+}
```
natebosch: [nit] I think it is more straightforward to create a suffix with the right length than to create a fixed length and shorten it.

```suggestion
String _randomSuffix(int length) => [
  for (int i = 0; i < length; i++)
    _bucketChars[_random.nextInt(_bucketChars.length)],
].join();

String _bucketName(String name) {
  assert(name.length < 55, '"$name" is too long to append a random suffix.');
  return TestHttpClient.isRecording || TestHttpClient.isReplaying
      ? name
      : '$name-${_randomSuffix(59 - name.length)}';
}
```

PR #143, opened 2026-02-20
feat(gcs): modify object metadata

brianquinlan:
1. Adds `Storage.patchObject`, which uses the new `ObjectMetadataPatchBuilder` to build a set of changes to the object.
3. Add the ability to enable object retention in `Storage.createBucket`
4. Add the ability to list all object versions in `Storage.listObjects`
5. Modify the test teardown to remove holds on objects so that they can be deleted and to delete all object versions
6. Adds tests for 1-5


brianquinlan:
> Can you update the opening comment in this PR with content for the commit message when we merge?

Done.

PR #155, opened 2026-02-26
feat(gcs): make http client optional

### LINKED ISSUE #149: gcs: Support configuring the API endpoint using STORAGE_EMULATOR_HOST
The GCS client should read the STORAGE_EMULATOR_HOST environment variable to determine the API endpoint unless one is implicitly set.

See the [Golang documentation](https://docs.cloud.google.com/go/docs/reference/cloud.google.com/go/storage/latest#hdr-Creating_a_Client) or [Python implementation](https://github.com/googleapis/python-storage/blob/d8dd1e074d2431de9b45e0103181dce749a447a0/google/cloud/storage/client.py#L194).

### LINKED ISSUE #151: gcs: Use application default credentials by default

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/149
Fixes https://github.com/googleapis/google-cloud-dart/issues/151

Makes the `http.Client` arguments to `Storage` optional. If the `http.Client` argument is not provided then a client that uses application default clients is created **unless** the `'STORAGE_EMULATOR_HOST'` environment variable is set. If that environment variable is set then an unauthenticated client is used and, unless an API endpoint is explicitly set, the endpoint set in the environment variable is used.

I also considered encapsulating the `FutureOr` fields in a helper class:
https://github.com/googleapis/google-cloud-dart/pull/155/changes/c1f2b21e79511cb4159cd490fc30e3734ce5979c


packages/google_cloud_storage/lib/src/client.dart
```diff
@@ -43,11 +45,45 @@ class _JsonEncodableWrapper implements JsonEncodable {
...
+
+  static String? _getStorageEmulatorHost() =>
+      Platform.environment['STORAGE_EMULATOR_HOST'];
+
+  static FutureOr<http.Client> _calculateClient(
+    http.Client? client,
+    String? emulatorHost,
+  ) => switch ((client, emulatorHost)) {
+    (final http.Client client, _) => client,
+    (null, final String _) => http.Client(),
```
natebosch: If you decide not to apply the other change, we do want to drop the `final`, we aren't using the declared variable so it doesn't need to be declared. The way to express a pattern which is solely a type check (what's be accomplished with `final String _`) is with `TypeName()`.


```suggestion
    (null, String()) => http.Client(),
```

natebosch (approved):

PR #217, opened 2026-03-26
Ignore vscode config

sigurdm:


brianquinlan (approved):

kevmoo (commented):
I'm a BIG fan of telling folks to put this in `~/.gitconfig`!!!

so this is a polity 👎 – but not a huge deal 🤷

PR #219, opened 2026-03-26
feat(storage): add support for streaming uploads

### LINKED ISSUE #180: add streaming uploads
The user should be able to stream object uploads using `IOSink` or similar interface

https://docs.cloud.google.com/storage/docs/streaming-uploads#rest-apis

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/180

Supports uploading GCS objects using a `StreamSink<List<int>>`.

This PR trades performance and correctness for simplicity. Future PRs (see the TODO in the code) will fix the correctness problems.

Calls to `add` accumulate data in memory until `close` is called, which causes the data to be sent to GCS.
Calls to `addStream` accumulate data into 256KiB chunks (the minimum GCS chunk size) and then sends that data to GCS.

It is possible for GCS to partially reject a write. For example, you could send 100MiB to GCS in a single HTTP request and GCS could indicate that 50MiB was successfully stored. That scenario is not currently supported and the upload will fail in that case.


pkgs/google_cloud_storage/lib/src/client.dart
```diff
@@ -790,6 +791,41 @@ final class Storage {
...
+  ///   'hello.txt',
+  /// );
+  /// sink.add([1, 2, 3]);
+  /// await sink.close();
+  /// ```
+  ///
+  /// See [API reference docs](https://docs.cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload).
+  ///
+  /// [Google Cloud Storage object]: https://docs.cloud.google.com/storage/docs/json_api/v1/objects
+  ResumableUploadSink uploadObjectFromSink(
```
natebosch: [nit] I don't think we need to expose the `ResumableUploadSink` type, I don't see any API surface it adds.


```suggestion
  StreamSink<List<int>> uploadObjectFromSink(
```

natebosch: Should we take a `Stream` argument instead of returning a `StreamSink`? It would let us return a `Future<ObjectMetadata>` like the other method.

natebosch (approved):

PR #230, opened 2026-04-07
Reworks the v0.4.1 release to be more sane

kevmoo:
- Aligns the behavior of logging when one is on and not on Google Cloud
- Finalizes the stand-alone logger name to errorLoggingMiddleware
- errorLoggingMiddleware now handles all exception uniformly and WRT JSON responses

### DIFF
--- a/pkgs/google_cloud/lib/src/serving/http_logging.dart
+++ b/pkgs/google_cloud/lib/src/serving/http_logging.dart
@@ -106,7 +106,7 @@ Handler _handleResponseException(Handler innerHandler) {
   return logRequests()(exceptionHandler(innerHandler));
 }
 
-/// Creates a [Response] from an [error] and [stack].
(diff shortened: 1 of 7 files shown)


pkgs/google_cloud/lib/src/serving/http_logging.dart
```diff
@@ -28,92 +28,136 @@ import 'trace_context_data.dart';
...
+      ].expand((e) => LineSplitter.split('$e'.trim())).join('\n');
+
+      stderr.writeln(text);
+      return _responseFromException(error, stack, request.headers);
+    }
+  };
+  return logRequests()(exceptionHandler(innerHandler));
+}
 
+/// Creates a [Response] from an [error] and [stack].
```
brianquinlan:
```suggestion
/// Creates a [Response] from [error] and [stack].
```

brianquinlan (approved):

PR #232, opened 2026-04-07
feat(storage): generate hashes for sink upload

brianquinlan:


pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -60,6 +67,12 @@ class ResumableUploadSink implements StreamSink<List<int>> {
   }
 
   void _addToBuffer(List<int> data) {
+    if (data.isEmpty) {
+      return;
+    }
```
natebosch: [optional] I'm a fan of omitting braces for guard clause conditionals.

```suggestion
    if (data.isEmpty) return;
```

natebosch (approved):

PR #234, opened 2026-04-08
feat(storage): return metadata in sink upload

brianquinlan:


pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -37,8 +37,16 @@ int _largestWriteSize(int number) => (number ~/ _minWriteSize) * _minWriteSize;
 class ResumableUploadSink implements StreamSink<List<int>> {
   bool _isClosing = false;
   bool _isAddStream = false;
-  final Completer<bool> _closedCompleter = Completer<bool>();
+  final Completer<ObjectMetadata> _closedCompleter =
+      Completer<ObjectMetadata>();
```
natebosch: [optional] It's OK to omit the type on the left when the type on the right is obvious.


```suggestion
  final _closedCompleter = Completer<ObjectMetadata>();
```

natebosch (approved):

PR #235, opened 2026-04-09
test(storage): run sink tests using testbench

brianquinlan:
A follow-up PR will introduce some interesting new tests.

### DIFF
--- a/pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
+++ b/pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
@@ -35,13 +35,13 @@ Uint8List randomUint8List(int length, {int? seed}) {
   return l;
 }
 
-void uploadObjectFromSinkTest(Storage Function() storageFn) {
+void uploadObjectFromSinkTest(Storage Function() createStorage) {
   late Storage storage;
   final small = randomUint8List(100);
   final large = randomUint8List(5_000_000);
 
   setUp(() {
-    storage = storageFn();
+    storage = createStorage();
   });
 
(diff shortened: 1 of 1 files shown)


pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
```diff
@@ -35,353 +35,367 @@ Uint8List randomUint8List(int length, {int? seed}) {
   return l;
 }
 
-void main() async {
+void uploadObjectFromSinkTest(Storage Function() storageFn) {
```
natebosch: [nit] Prefer to avoid abbreviations like "Fn". Maybe `Storage Function() createStorage`?

pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
```diff
@@ -35,353 +35,367 @@ Uint8List randomUint8List(int length, {int? seed}) {
...
+        [10, 11, 12],
+      ]),
+    );
+    sink
+      ..add([13, 14, 15])
+      ..add([16, 17, 18]);
+    await sink.close();
+
+    final downloaded = await storage.downloadObject(bucketName, 'name');
+    expect(downloaded, List.generate(18, (i) => i + 1));
```
natebosch: [out of scope] I probably overlooked this before, but I generally prefer list literals over `List.generate`.

```suggestion
    expect(downloaded, [for (var i = 0; i < 18; i++) i + 1]);
```

natebosch (approved):

PR #236, opened 2026-04-14
feat(storage): support retries with sink uploads

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/218

Adds `ifGenerationMatch` and `retry` parameters to uploadObjectFromSink.

Part of the complexity of the retry strategy is that it is possible for Google Cloud Storage to only acknowledge part of the upload **and** uploads must be done in 256KB chunks.


pkgs/google_cloud_storage/lib/src/client.dart
```diff
@@ -783,10 +783,10 @@ final class Storage {
   /// If [metadata] is non-null, it will be used as the object's metadata. If
   /// `metadata.name` does not match [name], a [BadRequestException] is thrown.
   ///
-  /// If set, `ifGenerationMatch` makes updating the object content conditional
+  /// If set, [ifGenerationMatch] makes updating the object content conditional
   /// on whether the object's generation matches the provided value. If the
   /// generation does not match, a [PreconditionFailedException] is thrown.
-  /// A value of `0` indicates that the object must not already exist.
+  /// A value of `BigInt.zero` indicates that the object must not already exist.
```
natebosch: [optional] I think it should work to link this reference.


```suggestion
  /// A value of [BigInt.zero] indicates that the object must not already exist.
```

pkgs/google_cloud_storage/lib/src/client.dart
```diff
@@ -854,17 +854,25 @@ final class Storage {
...
+  ///
   /// If [metadata] is non-null, it will be used as the object's metadata. If
   /// [metadata] is `null` or `metadata.contentType` is `null`, the content type
   /// will be `'application/octet-stream'`. If `metadata.name` does not
   /// match [name], a [BadRequestException] is thrown.
   ///
+  /// If set, [ifGenerationMatch] makes updating the object content conditional
+  /// on whether the object's generation matches the provided value. If the
+  /// generation does not match, a [PreconditionFailedException] is thrown.
+  /// A value of `BigInt.zero` indicates that the object must not already exist.
```
natebosch: [optional]

```suggestion
  /// A value of [BigInt.zero] indicates that the object must not already exist.
```

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -33,12 +34,31 @@ const _minWriteSize = 256 * 1024;
...
+///
+/// From https://docs.cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload
+///
+/// > Repeat the above steps for each remaining chunk of data that you want
+/// > to upload, using the upper value contained in the Range header of each
+/// > response to determine where to start each successive chunk; you should
+/// > not assume that the server received all bytes sent in any given request.
+int? _parseRange(String? rangeHeader) {
+  if (rangeHeader == null) return null;
+  final match = RegExp(r'bytes=0-(\d+)').firstMatch(rangeHeader);
```
natebosch, line 48: [optional] Gemini tells me that GCS does always use `0-` ranges (I didn't look it up to confirm), but I wonder it it makes sense to be target the more general spec anyway?


```suggestion
  final match = RegExp(r'bytes=\d+-(\d+)').firstMatch(rangeHeader);
```

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -33,12 +34,31 @@ const _minWriteSize = 256 * 1024;
...
+  if (match != null) {
+    return int.parse(match.group(1)!) + 1;
+  }
```
natebosch: I like to avoid the `!` when we can.

```suggestion
  if (match?.group(1) case final group?) {
    return int.parse(group) + 1;
  }
```

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -77,53 +96,144 @@ class ResumableUploadSink implements StreamSink<List<int>> {
...
-      var newSize = _writeBuffer.length * 2;
-      if (newSize < requiredCapacity) {
-        newSize = requiredCapacity;
+  Future<http.Response> _uploadChunk(
+    Uint8List chunk,
+    bool isClose,
+    String? hashHeader,
+  ) async {
+    final sessionUri = await _sessionUri;
+    final client = await _client;
```
natebosch, line 108: [optional] It doesn't make a difference here, but since I went looking to check, I wonder if it doesn't make sense to use the parallel wait API so future readers don't go looking for the same?


```suggestion
    final (sessionUri, client) = await (_sessionUri, _client).wait;
```

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -77,53 +96,144 @@ class ResumableUploadSink implements StreamSink<List<int>> {
...
+            headers: {'Content-Range': 'bytes */*'},
+          );
+          if (statusRes.statusCode == 308) {
+            loopExpectedByte = _parseRange(statusRes.headers['range']) ?? 0;
+          } else if (statusRes.statusCode == 200 ||
+              statusRes.statusCode == 201) {
+            return statusRes;
+          } else {
+            throw ServiceException.fromHttpResponse(statusRes, statusRes.body);
+          }
```
natebosch: [optional] Since 2 of 3 conditions in the chain leave the function I think it's  bit easer to read as guard clauses.

```suggestion
          if (statusRes.statusCode == 200 || statusRes.statusCode == 201) {
            return statusRes;
          }
          if (statusRes.statusCOde != 308) {
            throw ServiceException.fromHttpResponse(statusRes, statusRes.body);
          }
          loopExpectedByte = _parseRange(statusRes.headers['range']) ?? 0;
```

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -77,53 +96,144 @@ class ResumableUploadSink implements StreamSink<List<int>> {
...
+  }
 
-    final requiredCapacity = _writeBufferSize + data.length;
-    if (requiredCapacity > _writeBuffer.length) {
-      var newSize = _writeBuffer.length * 2;
-      if (newSize < requiredCapacity) {
-        newSize = requiredCapacity;
+  Future<http.Response> _uploadChunk(
+    Uint8List chunk,
+    bool isClose,
```
natebosch: [nit] What is the specific meaning of "close" here? `isClosed`? `isClosingChunk`? `isNearby`?

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -217,6 +315,8 @@ ResumableUploadSink uploadFileStream(
   FutureOr<http.Client> client,
   Uri url, {
   ObjectMetadata? metadata,
+  bool isIdempotent = false,
```
natebosch: [nit] Can you add a doc comment on the method that describes how to use this argument?

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -230,16 +330,18 @@ ResumableUploadSink uploadFileStream(
...
-        'Content-Type': 'application/json',
-        'Content-Length': body.length.toString(),
-      },
-    ),
-  );
+      headers: {'Content-Type': 'application/json'},
+    );
+    if (res.statusCode < 200 || res.statusCode >= 300) {
+      throw ServiceException.fromHttpResponse(res, res.body);
+    }
```
natebosch, line 345: [optional] I like how patterns read here.

```suggestion
    if (res.statusCode case < 200 || >= 300) {
      throw ServiceException.fromHttpResponse(res, res.body);
    }
```

pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
```diff
@@ -431,9 +576,155 @@ void main() {
...
+          // Status check returns a range smaller than what was acknowledged
+          // previously.
+          return http.Response('', 308, headers: {'range': 'bytes=0-1000'});
+        } else {
+          throw StateError(
+            'Unexpected call (count: $count, method: ${request.method}, '
+            'uri: ${request.url})',
+          );
+        }
+      });
```
natebosch, line 642: [optional] The `count` behavior, especially in the second tests where it's used in an expectation, feels a little more indirect than I like for tests.

What about something like a `List<(String, http.Response Function()>` like

```dart
final responses = <(String, http.Response Function())>[
  ('POST', () => http.Response('', 200, headers: {'location': 'http://example.com/location'})),
  ('PUT',  () => http.Response('', 308, headers: {'range': 'bytes=0-262143'})),
  ('PUT',  () => throw http.ClientException('message')),
  ('PUT',  () => http.Response('', 308, headers: {'range': 'bytes=0-1000'})),
];
```

Use `responses.removeAt(0)` to know what method to check for, and use the callback to respond. The `expect(count, 6)` can become something like `expect(responses, isEmpty);`, which doesn't carry all the semantics of the test in the expectation, but I think will be a little more clear than a failure like "expected: 6, actual: 5"

natebosch (approved):

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
brianquinlan: This header is GCS' acknowledgment of the bytes that it has received. The documentation doesn't describe how to interpret the lower value

> Repeat the above steps for each remaining chunk of data that you want to upload, using the upper value contained in the Range header of each response to determine where to start each successive chunk; you should not assume that the server received all bytes sent in any given request.

If the lower value is non-zero then I treat it like the "Range" header was absent, which means that:
1. in a check request, assume that all bytes must be resent
2. in an upload request, assume that the state of the upload is unknown and do a check request

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
brianquinlan: I changed this to `isLast` with a note that this is the final chunk to upload.

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
brianquinlan: I didn't follow your exact advice but I restructured this a bit to be more (IMO) readable.

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
brianquinlan: Minor disagree on that one ;-)

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
brianquinlan: Done.

pkgs/google_cloud_storage/test/upload_object_from_sink_test.dart
brianquinlan: Done. I'm neutral on whether this is better.

PR #239, opened 2026-04-15
chore(cloud): downgrade meta to `^1.17.0`

brianquinlan:
Part of https://github.com/firebase/firebase-functions-dart/issues/133


kevmoo (commented):
I need to validate that we don't break `firebase_functions` with this.

I forgot that I had these deprecations in here!

kevmoo (approved):

PR #243, opened 2026-04-17
chore(cloud): revert rollback of 0.4.1 changes

brianquinlan:
Rollback https://github.com/googleapis/google-cloud-dart/commit/0094d0005480e02c05cae8ef599137abc9e57cdd but:
- Create a new version number (0.4.2-wip)
- Preserve the usage of `meta: ^1.17.0`


kevmoo (approved):
So this gets us back to where we WERE? Cool!

I'm guessing we'll move most of this delta to the other packages. But at least it'll be in the same repo w/ some git history, etc.

PR #257, opened 2026-04-30
fix(protobuf): fix `FieldMask` encoding

### LINKED ISSUE #253: FieldMask not correctly serialized using camelCase
From https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/field_mask.proto

```
// In JSON, a field mask is encoded as a single string where paths are
// separated by a comma. Fields name in each path are converted
// to/from lower-camel naming conventions.

```

brianquinlan:
Fixes https://github.com/googleapis/google-cloud-dart/issues/253

### DIFF
--- a/generated/google_cloud_protobuf/lib/src/protobuf.p.dart
+++ b/generated/google_cloud_protobuf/lib/src/protobuf.p.dart
@@ -310,11 +310,65 @@ class _DurationHelper {
 
 class _FieldMaskHelper {
   /// Encode the field mask as a single comma-separated string.
-  static String encode(FieldMask fieldMask) => fieldMask.paths.join(',');
+  ///
+  /// See https://protobuf.dev/reference/php/api-docs/Google/Protobuf/FieldMask.html
+  /// See https://github.com/protocolbuffers/protobuf/blob/d4ea3dd49c093435443c129497a3e1f0d5b8cd3e/java/util/src/main/java/com/google/protobuf/util/FieldMaskUtil.java#L137
+  static String encode(FieldMask fieldMask) {
+    final paths = <String>[];
+    for (final path in fieldMask.paths) {
+      if (path.isNotEmpty) {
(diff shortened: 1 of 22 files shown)


generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -310,11 +310,65 @@ class _DurationHelper {
...
+  /// See https://github.com/protocolbuffers/protobuf/blob/d4ea3dd49c093435443c129497a3e1f0d5b8cd3e/java/util/src/main/java/com/google/protobuf/util/FieldMaskUtil.java#L137
+  static String encode(FieldMask fieldMask) {
+    final paths = <String>[];
+    for (final path in fieldMask.paths) {
+      if (path.isNotEmpty) {
+        paths.add(_lowerUnderscoreToLowerCamel(path));
+      }
+    }
+    return paths.join(',');
+  }
```
natebosch:
```suggestion
  static String encode(FieldMask fieldMask) => fieldMask.paths
      .where((path) => path.isNotEmpty)
      .map(_lowerUnderscoreToLowerCamel)
      .join(',');
```

generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -310,11 +310,65 @@ class _DurationHelper {
...
+      throw FormatException(
+        'Expected string for FieldMask, got ${format.runtimeType}',
+      );
+    }
+    final paths = <String>[];
+    for (final path in format.split(',')) {
+      if (path.isNotEmpty) {
+        paths.add(_lowerCamelToLowerUnderscore(path));
+      }
+    }
```
natebosch:
```suggestion
    final paths = <String>[
      for (final path in format.split(','))
        if (path.isNotEmpty) _lowerCamelToLowerUnderscore(path),
    ];
```

natebosch (approved):

PR #259, opened 2026-05-06
chore(deps_diagram): Move to mermaid

kevmoo:
- Drop the `deps.png` file (differences in dot version may cause deltas that are hard to fix!)
- Embed the data the markdown!


kevmoo:
image

brianquinlan (commented):
That looks cool! Is there a way to force the layering? In the current diagram "google_cloud_protobuf" and "google_cloud" are on the same level because they are not dependent on each other. In the mermaid diagram, "google_cloud" is one level up.

I originally created this diagram to visualize the dependencies for publishing and that is harder now.

brianquinlan (approved):

PR #267, opened 2026-05-08
chore: add publish github workflow

kevmoo:


brianquinlan (approved):
So do you tag the release with the package name and then the version number. For example: `git tag google_cloud-v0.4.1`

If so, do we need to include _ in the first pattern?

Would you add a section in the developer guide covering how to use this?

kevmoo:
> So do you tag the release with the package name and then the version number. For example: `git tag google_cloud-v0.4.1`
> 
> If so, do we need to include _ in the first pattern?
> 
> Would you add a section in the developer guide covering how to use this?

This is the BEAUTY! The tool gives you the link to do the tagging. You click on it and you go to github's release UI, which creates the tag.

THEN the publish action takes over and publishes to github!

kevmoo:
dev guide details added!

PR #270, opened 2026-05-10
feat(storage): add all object acl methods

brianquinlan:
Add these methods to `Storage`:
- `getObjectAcl`
- `listObjectAcl`
- `patchObjectAcl`
- `updateObjectAcl`

### DIFF
--- a/pkgs/google_cloud_storage/test/patch_object_acl_test.dart
+++ b/pkgs/google_cloud_storage/test/patch_object_acl_test.dart
@@ -1,0 +1,238 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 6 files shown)


pkgs/google_cloud_storage/test/patch_object_acl_test.dart
```diff
@@ -0,0 +1,238 @@
...
+      final acl = await storage.patchObjectAcl(
+        'bucket',
+        'object',
+        'user-test',
+        'OWNER',
+        generation: BigInt.from(1),
+      );
+      expect(acl.entity, 'user-test');
+      expect(acl.role, 'OWNER');
+      expect(count, 2);
```
natebosch, line 235: [optional]
The `package:test` idiom for tracking whether a callback is called is `expectAsync*`.
https://pub.dev/documentation/test/latest/test/expectAsync1.html

```dart
final mockClient = MockClient(
  expectAsync1(
    count: 2,
    (request) async {},
  )
 );

natebosch (approved):

pkgs/google_cloud_storage/test/patch_object_acl_test.dart
natebosch: BTW this is an API we still need to port to `package:checks` so if you have any ideas on alternative designs I'd be happy to see them.

brianquinlan: I've used the count assert in my other PRs so I'll keep it for consistency and then fix it in a follow-up (or not - I'd want to check that the extra level of nesting doesn't make things look worse, also I need to track `count` in any case).

natebosch: > also I need to track `count` in any case

That's true, since we anyway need to distinguish each call it doesn't save any complexity to use the utility.

PR #280, opened 2026-05-19
feat(logging): `package:logging` integration

brianquinlan:
Allows this:

```dart
import 'package:google_cloud_logging/google_cloud_logging.dart';
import 'package:logging/logging.dart';

void main() {
  const cloudLogger = CloudLogger.structuredLogger();
  Logger.root.onRecord.listen(cloudLogger.handleLog);
  Logger.root.level = Level.ALL;

  Logger('MyClassName').fine('Starting file copy.');
}
```

### DIFF
--- a/pkgs/google_cloud_logging/CHANGELOG.md
+++ b/pkgs/google_cloud_logging/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.1
+## 0.5.1-wip
 
 - Add integrations for `package:logging`.
 
--- a/pkgs/google_cloud_logging/CHANGELOG.md
+++ b/pkgs/google_cloud_logging/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.1
+
+- Add integrations for `package:logging`.
+
 ## 0.5.0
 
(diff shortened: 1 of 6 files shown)


kevmoo (commented):
Do we think this is high priority? I'd rather wait until someone says they WANT this.

natebosch:
> Do we think this is high priority? I'd rather wait until someone says they WANT this.

+1 The implementation here looks good but I would rather wait to land it when we have a concrete use case.

brianquinlan:
> Do we think this is high priority? I'd rather wait until someone says they WANT this.

I implemented this for two reasons:
1. If developers use `CloudLogger` directly then they are tying themselves unnecessarily to Google Cloud. I'd rather that they use a standard logging package.
2. There are 1,688 packages on pub that depend on `package:logging` (I'd also support `package:logger` since there are 980 packages that use it). IIUC, none of the output of those packages will have their metadata (most importantly severity) searchable by the GCP Logs Explorer if they use the standard log handler.

image

brianquinlan:
This is now redundant.

PR #283, opened 2026-05-20
chore(logging): refactor for trace correlation

brianquinlan:
The most important ideas in this change are:
1. make it so `package:google_cloud_logging` does automatic trace correlation when `package:google_cloud_shelf` middleware is used.
2. make it so `package:google_cloud_logging` lots are not double encapsulated when used inside a `package:google_cloud_shelf` request

Implementation:
1. `package:google_cloud_shelf` no longer has exports a logger, instead it sets `Zone` variables that contain trace context that can be used by `package:google_cloud_logging`
2. `CloudLogger` has been removed and replaced by a single class `StructuredLogger`
3. removed `createStructuredLogFromEntry`, which is no longer needed
4. switched to using the w3 standardized `'traceparent'` header

Built on: https://github.com/googleapis/google-cloud-dart/pull/286


pkgs/google_cloud_logging/lib/src/structured_logger.dart
```diff
@@ -0,0 +1,179 @@
...
+import 'package:logging/logging.dart' as logging;
+
+import 'structured_logging.dart' show createStructuredLog;
+import 'write_web.dart' if (dart.library.io) 'write_io.dart';
+
+/// A simple logger that outputs logging messages using [Structured logging][1].
+///
+/// ```dart
+/// import 'package:google_cloud_logging/google_cloud_logging.dart';
+/// const _logger = StructuredLogger();
```
natebosch, line 26: [nit]
```suggestion
/// import 'package:google_cloud_logging/google_cloud_logging.dart';
///
/// const _logger = StructuredLogger();
```

pkgs/google_cloud_logging/lib/src/structured_logging.dart
```diff
@@ -12,156 +12,109 @@
...
-// Special fields that need to be mapped to google cloud format
-//
-// https://docs.cloud.google.com/logging/docs/structured-logging
-const _specialFieldsMapping = {
-  'trace': 'logging.googleapis.com/trace',
-  'spanId': 'logging.googleapis.com/spanId',
-  'traceSampled': 'logging.googleapis.com/trace_sampled',
-  'sourceLocation': 'logging.googleapis.com/sourceLocation',
-  'textPayload': 'message',
+Map<String, Object?> _filter(Map<dynamic, dynamic> m) => {
```
natebosch: [nit] add a doc and/or make the name more specific.

pkgs/google_cloud_logging/lib/src/structured_logging.dart
```diff
@@ -12,156 +12,109 @@
...
+    ...(traceparent == null
+        ? structuredTraceFromZone(projectId, zone)
+        : formatTraceparent(projectId, traceparent)),
+  };
+
+  if (stackTrace case final stackTrace?) {
+    result['stack_trace'] = _formatStackTrace(stackTrace).toString();
+    result['logging.googleapis.com/sourceLocation'] = _sourceLocation(
+      stackTrace,
     );
```
natebosch:
```suggestion
      if (stackTrace case final stackTrace?) ...{
        'stack_trace': _formatStackTrace(stackTrace).toString(),
        'logging.googleapis.com/sourceLocation': _sourceLocation(stackTrace),
      },
    };
```

pkgs/google_cloud_logging/lib/src/structured_logging.dart
```diff
@@ -12,156 +12,109 @@
...
   if (value == null || value is num || value is String || value is bool) {
     return value;
   }
-  seen ??= <Object>{};
+  seen ??= HashSet.identity();
   if (seen.contains(value)) {
-    throw const FormatException('Cyclic reference');
+    return '[CIRCULAR]';
   }
   seen.add(value);
```
natebosch:
```suggestion
  seen ??= HashSet.identity();
  if (!seen.add(value)) {
    return '[CIRCULAR]';
  }
```

The github suggestion is rendering weird - but we can take advantage of the `bool` return from `Set.add`.
https://api.dart.dev/dart-core/Set/add.html

pkgs/google_cloud_logging/lib/src/traceparent.dart
```diff
@@ -0,0 +1,99 @@
...
+final _traceParentRegex = RegExp(
+  '$_version-$_trace-$_parent-$_flags',
+  caseSensitive: false,
+);
+
+/// Parsers a `'traceparent'` header.
+///
+/// See https://www.w3.org/TR/trace-context/
+@visibleForTesting
+({String traceId, String spanId, bool traceSampled})? parseTraceparent(
```
natebosch, line 57: [nit] 
```suggestion
({String traceId, String spanId, bool traceSampled})? parseTraceParent(
```

pkgs/google_cloud_logging/test/traceparent_test.dart
```diff
@@ -0,0 +1,162 @@
+// Copyright 2026 Google LLC
```
natebosch, line 1: trace_parent_test?

The PR is a little inconsistent whether it's 1 word "traceparent" or 2 words "trace parent".

natebosch (approved):

pkgs/google_cloud_logging/lib/src/structured_logger.dart
```diff
@@ -0,0 +1,179 @@
...
+/// $ gcloud logging read 'jsonPayload.itemId = "A-987"'
+/// ---
+/// insertId: 654sjnf4ovq2l
+/// jsonPayload:
+///   itemId: A-987
+///   message: Processing item.
+/// ...
+/// ```
+/// When used with `package:google_cloud_shelf`, logs written using this logger
+/// are automatically associated with the request that generated them.
```
natebosch, line 46: Is it worth documenting the specific zone variables to define to make this work with third party serving frameworks?

kevmoo: I'd almost argue we should publish the zone variables as constants or something...

pkgs/google_cloud_logging/lib/src/structured_logging.dart
brianquinlan: That's cool!

pkgs/google_cloud_logging/lib/src/structured_logger.dart
brianquinlan: How about I make a new file `/lib/interop.dart` that exports the constants, has a link to the `ARCHITECTURE.md` doc and explains in general how this works?

That way end developers who shouldn't have to care about this can ignore it but framework developers have a good reference.

kevmoo (changes requested):
Give me till noon to review this! 😄

pkgs/google_cloud_logging/test/traceparent_test.dart
brianquinlan: The w3 standard makes it look like a single word. For example, they never put a space/hyphen between the words and capitalize it like "Traceparent". I couldn't find anywhere that I spelled it like it should be two words but I could have easily missed something.

pkgs/google_cloud_logging/lib/src/structured_logging.dart
brianquinlan: I'm a bit skeptical of this one - I had to look up the return value of `add`...but done.

pkgs/google_cloud_logging/lib/src/structured_logging.dart
brianquinlan: Done.

pkgs/google_cloud_logging/test/traceparent_test.dart
natebosch: Ah, I was overly anchored on the one place with difference capitalization because I had looked at it more closely. `_traceParentRegex` is the only place where it looks like 2 words.

pkgs/google_cloud_logging/lib/src/structured_logger.dart
natebosch: Separate library SGTM.

pkgs/google_cloud_logging/pubspec.yaml
```diff
@@ -13,9 +13,11 @@ environment:
 dependencies:
   google_cloud_logging_type: ^0.5.2
   google_cloud_logging_v2: ^0.5.2
-  google_cloud_protobuf: ^0.5.2
+  logger: ^2.7.0
```
kevmoo: We're depending on a 3p package? https://pub.dev/packages/logger ??

pkgs/google_cloud_logging/lib/src/structured_logger.dart
```diff
@@ -0,0 +1,180 @@
...
+      severity,
+      payload: extra,
+      zone: record.zone,
+      stackTrace: record.stackTrace,
+    );
+    _print(logStr);
+  }
+
+  logging_type.LogSeverity _severityFromLoggingLevel(logging.Level level) {
+    if (level <= logging.Level.FINE) {
```
kevmoo: I feel there is some prettier `switch/case` thingy one could do here, but I can't think of it off the top o' me 'ead.

kevmoo (dismissed):
@brianquinlan – did you completely get rid of just plain text printing (non-JSON?)

pkgs/google_cloud_logging/pubspec.yaml
brianquinlan: Oops. Removed (but hope to re-add support for that package later ;-))

pkgs/google_cloud_logging/lib/src/structured_logger.dart
natebosch:
```dart
    return switch (level) {
      <= logging.Level.FINE => logging_type.LogSeverity.debug,

      <= logging.Level.INFO => logging_type.LogSeverity.info,
      <= logging.Level.WARNING => logging_type.LogSeverity.warning,
      <= logging.Level.SEVERE => logging_type.LogSeverity.error,
      <= logging.Level.SHOUT => logging_type.LogSeverity.critical,
      _ => logging_type.LogSeverity.emergency,
    };
  }
```

PR #286, opened 2026-05-22
test(shelf): added integration tests for `print`

brianquinlan:
Added an integration test that verifies that calling the `print` function in the context of a request results in a log entry.

The test works by running a shelf server using Google Cloud Run and then verifying that the correct logs written.

1. Changed the GCB image to one that includes `gcloud`
2. Changed the integration test install script to install dart
3. Added `waitForLogs` to `test_utils` (logs are ingested asynchronously)
4. Added a `CloudRunner` class to run a Dart program using Google Cloud Run.

### DIFF
--- a/pkgs/google_cloud_shelf/test/logging_test.dart
+++ b/pkgs/google_cloud_shelf/test/logging_test.dart
@@ -232,12 +232,19 @@ Details:
     test('print', () async {
       final uniqueId = 'Hello World: ${randomAlphaNumString(20)}';
 
+      final headers = <String, String>{};
+      final token = await runner.getIdToken();
+      if (token != null) {
+        headers['Authorization'] = 'Bearer $token';
+      }
+
       // Trigger a print.
       final response = await http.get(
         runner.serverUri.replace(
           path: '/print',
           queryParameters: {'msg': uniqueId},
         ),
+        headers: headers,
       );
       expect(response.statusCode, 200);
 
--- a/pkgs/google_cloud_shelf/test/logging_test.dart
+++ b/pkgs/google_cloud_shelf/test/logging_test.dart
@@ -20,9 +20,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:google_cloud_logging/google_cloud_logging.dart';
-import 'package:google_cloud_logging_v2/logging.dart';
 import 'package:google_cloud_shelf/google_cloud_shelf.dart';
-import 'package:googleapis_auth/auth_io.dart' as auth;
 import 'package:http/http.dart' as http;
 import 'package:shelf/shelf.dart';
 import 'package:test/test.dart';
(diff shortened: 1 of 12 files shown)


brianquinlan:
The GCB tests are failing due to a permission error. I will not submit until that is resolved.

pkgs/google_cloud_shelf/test/logging_test.dart
```diff
@@ -212,6 +215,49 @@ Details:
...
+    });
+
+    test('print', () async {
+      final uniqueId = 'Hello World: ${randomAlphaNumString(20)}';
+
+      final headers = <String, String>{};
+      final token = await runner.getIdToken();
+      if (token != null) {
+        headers['Authorization'] = 'Bearer $token';
+      }
```
natebosch:
```suggestion
      final headers = <String, String>{
        if (await runner.idToken case final token?)
          'Authorization': 'Bearer $token',
      };
```

test_utils/lib/src/cloud_runner.dart
```diff
@@ -0,0 +1,197 @@
...
+      '--region',
+      'us-central1',
+      '--quiet',
+      '--project',
+      projectId,
+    ]);
+  }
+
+  /// Gets an ID token for the service URL.
+  Future<String?> getIdToken() async {
```
natebosch: https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get
```suggestion
  Future<String?> get idToken async {
```

test_utils/lib/src/cloud.dart
```diff
@@ -25,3 +27,49 @@ const googleTestUser = 'daenerysstone.938939@gmail.com';
...
+///
+/// Waits up to [timeout] for matching log entries to be available.
+///
+/// [filter] must be in [Logging query language][1]. For example,
+/// `'textPayload:"Hello World"'`.
+///
+/// [1]: (https://docs.cloud.google.com/logging/docs/view/logging-query-language)
+Future<List<LogEntry>> waitForLogs(
+  String filter, {
+  int count = 10,
```
natebosch, line 49: consider renaming to pageSize and/or adding an explanation of the behavior in the doc. From the naming I expected this to behavior like a minimum number of log entries to wait for, but that's not what I'd guess the `pageSize` argument is doing below.

test_utils/lib/src/cloud_runner.dart
```diff
@@ -0,0 +1,197 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:io';
+import 'package:path/path.dart' as p;
+import '../random.dart';
+import 'cloud.dart';
```
natebosch, line 18:
```suggestion
import 'dart:io';

import 'package:path/path.dart' as p;

import '../random.dart';
import 'cloud.dart';
```

Luckily `dart format` will do this for us soon 😄 https://github.com/dart-lang/dart_style/pull/1843

test_utils/lib/src/cloud_runner.dart
```diff
@@ -0,0 +1,197 @@
...
+String get repoRoot {
+  var dir = Directory.current.absolute;
+  while (true) {
+    if (File(p.join(dir.path, 'librarian.yaml')).existsSync()) {
+      return dir.path;
+    }
+    final parent = dir.parent;
+    if (parent.path == dir.path) {
+      break;
+    }
```
natebosch, line 43: [optional] Consider omitting braces for short guard clause conditonals.


```suggestion
    if (parent.path == dir.path) break;
```

test_utils/lib/src/cloud_runner.dart
natebosch:
```suggestion
  /// A fresh ID token for the service URL.
  Future<String?> get idToken async {
```
https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

natebosch (approved):

pkgs/google_cloud_shelf/test/logging_test.dart
brianquinlan: I restructured the code so this is encapsulated by `CloudRunner`

test_utils/lib/src/cloud.dart
brianquinlan: Added an explanation.

test_utils/lib/src/cloud_runner.dart
brianquinlan: Done.

test_utils/lib/src/cloud_runner.dart
brianquinlan: Done.

test_utils/lib/src/cloud_runner.dart
brianquinlan: Done.

PR #289, opened 2026-06-16
chore: update librarian to v0.21.0

kevmoo:
Updates Librarian API generator and configurations to stable v0.21.0 release checkpoint and regenerates all API packages.


natebosch (approved):
Almost 2k net negative lines 🤩

PR #292, opened 2026-06-22
feat(google_cloud_pubsub): add Subscription ack and modifyAckDeadline batching

### LINKED ISSUE #346: Share retry logic between google_cloud_pubsub and google_cloud_storage (before pubsub 1.0)
`google_cloud_pubsub` and `google_cloud_storage` each have their own retry
implementation, and the `ServiceException` half of the retryability predicate
(408/429/500/502/503/504) is currently duplicated verbatim between them.

- `pkgs/google_cloud_pubsub/lib/src/retry.dart` — `RetrySettings` data class
  (validated, non-const, `==`/`hashCode`), plus `@internal` `delaySequence`,
  `isRetryable`, `runWithRetry`. Only `RetrySettings` is exported.
- `pkgs/google_cloud_storage/lib/src/retry.dart` — `sealed RetryRunner` /
  `ExponentialRetry` / `NoDelayRetry` / `const defaultRetry`, exported
  wholesale and threaded through ~39 public signatures.

Context: https://github.com/googleapis/google-cloud-dart/pull/292

## Suggested order

1. Share the **internals** first (`delaySequence`, `runWithRetry`, and the
   `ServiceException` arm of the predicate). This is invisible to users and
   needs no public API decision.
2. Only then consider unifying the public settings type.

## Open questions to resolve first

- **Home package.** `package:google_cloud_rpc` is the only package both
  already depend on (`^0.6.0`), but it is generated: `lib/rpc.dart` is not in
  librarian's `keep` list, so hand-written `lib/retry.dart` +
  `lib/src/retry.dart` + tests must each be added to `keep`, and `clock` added
  to its dependencies — which makes `clock` transitive for every generated
  REST package. It must also stay free of `package:grpc`.
- **`const` vs validation.** `defaultRetry` is a `const` default parameter
  value in ~39 storage signatures, so a merged type must have a `const`
  constructor — which cannot have a body. But `assert`s cannot express the
  `Duration` invariants at all (`assert(initialDelay > Duration.zero)` in a
  const constructor is a compile error: *"In constant expressions, operands of
  this operator must be of type 'num'"*), and the `int`/`double` asserts that
  *can* be written do not run under `dart run` or `dart compile exe`. So
  unifying the type means losing eager `ArgumentError` validation unless a
  separate `checkValid()` is introduced and called from every non-const
  boundary.
- **`NoDelayRetry`.** It is `const ... (initialDelay: Duration.zero,
  maxDelay: Duration.zero)`, which pubsub's validation explicitly rejects.
- **Jitter.** pubsub applies ±20%; storage applies none. Unifying changes
  storage's runtime retry timing.
- **Defaults must stay per-service.** pubsub 100 ms / ×1.3 / 1 min vs storage
  1 s / ×2 / 2 min. Do not share `defaultRetry`.
- **Naming.** pubsub's `totalTimeout` matches GAX; storage's
  `maxRetryInterval` reads like a per-attempt cap, which is actually
  `maxDelay`.

## Constraint

Move the declaration and re-export it — never introduce a second same-named
type. Two `RetrySettings` declarations would be an ambiguous import for anyone
using both packages.

Until this is done, any fix to retryability must be applied to both copies.

TAG=agy

sigurdm:
Implements request batching and exponential backoff retry infrastructure for `package:google_cloud_pubsub`.

> **Note:** To make this easier to review, this PR has been split into a 4-PR stack (and ~2,700 lines of duplicate test files/fakes have been consolidated into the existing per-feature test files):
> 1. **#352** — Extract shared `RetryRunner` and `ExponentialRetry` into `package:google_cloud_rpc` (+412 / -179 LOC, closes #346)
> 2. **#353** — Add `google_cloud_pubsub` retry configuration and error classification (+421 / -8 LOC)
> 3. **#354** — Add `Batcher` and `Topic` publish batching (+943 / -131 LOC)
> 4. **#355** — Add `Subscription` ack and `modifyAckDeadline` batching (+753 / -170 LOC)

Closes #346

TAG=agy


kevmoo:
Where are we here?

pkgs/google_cloud_pubsub/lib/src/retry.dart
```diff
@@ -0,0 +1,273 @@
...
+
+import 'dart:math';
+
+import 'package:clock/clock.dart';
+import 'package:google_cloud_rpc/exceptions.dart';
+import 'package:grpc/grpc.dart';
+import 'package:meta/meta.dart';
+
+/// Settings for configuring retry logic with exponential backoff.
+final class RetrySettings {
```
brianquinlan, line 27: In other language's ecosystem's, this goes elsewhere so that I can be reused across packages:
https://docs.cloud.google.com/pubsub/docs/retry-requests#retry_a_message_request

I guess that it doesn't cost us anything to have this here for now and then move it later (we can use `export` for backwards compatibility).

pkgs/google_cloud_pubsub/lib/src/batching.dart
```diff
@@ -0,0 +1,161 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+/// Settings for batching operations.
+final class BatchingSettings {
```
brianquinlan, line 20: Interesting, this is a generic class in Java:
https://docs.cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.batching.BatchingSettings.html

But PubSub specific in Python:
https://docs.cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.BatchSettings

Maybe add some of the documentation details from the Python version. `max_bytes` in particular seems helpful:

> The maximum total size of the messages to collect before automatically publishing the batch, including any byte size overhead of the publish request itself. The maximum value is bound by the server-side limit of 10_000_000 bytes. 

Or maybe the part about overhead is not true?

pkgs/google_cloud_pubsub/lib/src/retry.dart
```diff
@@ -0,0 +1,273 @@
...
+              nextMicroseconds >= maxDelay.inMicroseconds)
+          ? maxDelay
+          : Duration(microseconds: nextMicroseconds.round());
+    }
+  }
+}
+
+/// Returns whether [error] is considered a retryable error.
+@internal
+bool isRetryable(Object error) {
```
brianquinlan, line 191: Probably we can't share code with https://github.com/googleapis/google-cloud-dart/blob/main/pkgs/google_cloud_storage/lib/src/retry.dart , right?

pkgs/google_cloud_pubsub/lib/src/subscription.dart
```diff
@@ -45,8 +107,23 @@ final class Subscription {
...
+
+  void _initBatchers() {
+    _ackBatcher = Batcher<_AckRequest>(
+      settings: ackSettings.batching,
+      itemSize: (request) => request.ackId.length,
+      onBatch: _onAckBatch,
+    );
+    _modifyAckBatcher = Batcher<_ModifyAckDeadlineRequest>(
+      settings: ackSettings.batching,
+      itemSize: (request) => request.ackId.length + 4,
```
brianquinlan, line 127: Maybe add a comment explaining the `+4`?

brianquinlan (commented):
You didn't want to add any end-to-end tests?

pkgs/google_cloud_pubsub/lib/src/batching.dart
sigurdm: You were right to be suspicious, and it turned out to be worth chasing — thank you for pushing on it. The overhead sentence was **not** true for our implementation, and making it true uncovered a second, worse problem. Both are now fixed.

### The publish path

We were counting only `data.length` plus the UTF-8 length of each attribute key and value. The server, though, validates `PublishRequest.ByteSizeLong()` — the fully serialized request, including the topic field, the per-message tag and length prefix, and the attribute-map framing. So `maxBytes` was measuring something the server doesn't.

At the 1 MiB default this is harmless (0.06% overhead, and we sit 10× under the limit), but it is not harmless if you raise it. Measured:

| `maxBytes` | message shape | counted | real wire | vs 10,000,000 |
|---|---|---|---|---|
| 1 MiB (default) | 10 KiB | 1,024,000 | 1,024,663 | fine |
| 10,000,000 | 1 KiB | 9,999,360 | 10,058,013 | **+58 KB over** |
| 10,000,000 | 100 B | 10,000,000 | 10,400,063 | **+400 KB over** |
| 10,000,000 | 10 B + 5 attributes | 9,999,910 | 12,058,778 | **+2 MB over (20.6%)** |

An oversized request fails with a non-retryable `INVALID_ARGUMENT` and takes the whole batch with it.

Looking at the other clients, we were the outlier: Python and Java count the framing exactly, Go approximates it and clamps, Node counts naively like we did but clamps to 9 MiB for ~563 KB of headroom. We had neither. (Your instinct to quote Python's docstring was right — that wording *is* the ecosystem's contract.)

Fixed by counting wire sizes arithmetically without serializing (`package:protobuf` has no cached-size API). New `test/wire_size_test.dart` pins the arithmetic to real `writeToBuffer()` output across varint boundaries, empty payloads, empty attribute keys/values, multi-byte UTF-8, and 2,000 random messages.

### The acknowledgment path — the one I didn't expect

`AckSettings` shares `BatchingSettings` with publishing, so the ack and modack batchers inherited `maxBytes` = 1 MiB. The server limit for `Acknowledge` and `ModifyAckDeadline` is **512 KB**. The default was 2× the limit.

Nothing broke in practice only because `maxMessages: 100` always binds first. Raise `maxMessages` and you'd hit it — and because `_onAckBatch` swallows errors for fire-and-forget callers, the symptom would not be an exception but **every message in the batch being redelivered indefinitely**.

`AckSettings` now defaults `maxBytes` to 512 KB, and ack batches are measured with the same exact accounting.

### Also

- Settings above the applicable server limit now **throw** an `ArgumentError`
  rather than silently producing a doomed request. Python, Go and Node all
  clamp instead; clamping felt worse here, because
  `BatchingSettings(maxBytes: 50000000)` silently becoming 10,000,000 gives you
  no signal that you asked for something impossible.

  There is one wrinkle this created. `maxBytes` defaults to 1 MiB, which is
  right for publishing but above the 512,000-byte acknowledgment limit — so
  `AckSettings(batching: BatchingSettings(maxMessages: 50))` would throw over a
  byte limit the caller never expressed an opinion about. `BatchingSettings`
  therefore records privately whether `maxBytes` was passed explicitly, and only
  throws if it was; an unspecified default is quietly narrowed. The public field
  stays a non-nullable `int` with an unchanged default, so the API shape doesn't
  change. The alternative — making `maxBytes` an `int?` — is cleaner
  conceptually but makes "what is the default" unanswerable from the field, and
  validating in the constructor is impossible since the applicable limit depends
  on which RPC the settings end up feeding.
- `maxMessages` gets the same treatment against the 1,000-message publish limit.
  It needs no explicitness flag: its default of 100 is below every count limit,
  so anything over is necessarily deliberate.
- The `+4` in the modack batcher is now the exact varint size of the deadline. While fixing it I noticed the comment I'd just added was wrong: only `StreamingPullRequest` has the parallel deadline list; a unary `ModifyAckDeadlineRequest` has a single shared deadline. The comment says so now, and notes that this makes the estimate conservative on the unary path.

Regression tests on both paths: for publishing, 4,000 small messages with five attributes each (the worst case for framing), asserting that no request exceeds `maxBytes`; for acknowledgments, `test/subscription_batching_test.dart` measures every `AcknowledgeRequest` and `ModifyAckDeadlineRequest` emitted with `writeToBuffer()`, including batches that fan out across multiple distinct deadlines.

One thing I did **not** do: reject a single message too large to ever fit. Python raises `MessageTooLargeError` client-side for this; we still let the server reject it. Happy to add that if you'd like it.

pkgs/google_cloud_pubsub/lib/src/subscription.dart
sigurdm: Thanks for questioning this — looking into the `+4` made me realize it was
wrong, and fixing it properly meant fixing how the whole batch is measured.
It's gone now, replaced by exact accounting, so rather than explain the
constant I'll explain what replaced it.

Two things were wrong with `+4`:

1. **The deadline is a varint, not a fixed 4 bytes.** `int32` on the wire is
   base-128 varint encoded, so a typical 600-second deadline costs 2 bytes, not
   4. It's now `varintSize(request.ackDeadlineSeconds)`.
2. **My justification for it was wrong too.** I'd written that
   `ModifyAckDeadline` carries a `modifyDeadlineSeconds` list parallel to the
   ack ID list. That's only true of `StreamingPullRequest` — a unary
   `ModifyAckDeadlineRequest` has a *single shared* `ackDeadlineSeconds`
   (field 3). One batcher feeds both paths, so it still charges each ack ID for
   its own deadline, but the comment now says that this makes the estimate
   conservative on the unary path rather than exact.

More importantly, the old accounting counted only `ackId.length` and nothing
else — not the per-ack-ID field tag and length prefix, and not the subscription
name that every request carries. That mattered much more than the `+4`: see my
reply on `batching.dart` for the full story, but the short version is that
`AckSettings` was sharing publishing's 1 MiB `maxBytes` against a server limit
of 512,000 bytes for `Acknowledge` and `ModifyAckDeadline`.

The batcher now seeds its running total with the serialized size of the
subscription field and charges each ack ID its real framing, and
`test/subscription_batching_test.dart` measures every request the client
actually emits with `writeToBuffer()`. Reverting to the old accounting fails
three of those tests.

pkgs/google_cloud_pubsub/lib/src/retry.dart
sigurdm: Good call — since `google_cloud_pubsub` hasn't been published yet, I went ahead and moved this now rather than deferring. `RetryRunner` and `ExponentialRetry` now live in `package:google_cloud_rpc/retry.dart`, and both `google_cloud_storage` and `google_cloud_pubsub` re-export and use them (closes #346).

pkgs/google_cloud_pubsub/lib/src/retry.dart
sigurdm: We can! I unified both packages on `storage`'s `RetryRunner` / `ExponentialRetry` design in `package:google_cloud_rpc/retry.dart` (`google_cloud_storage` re-exports it for backwards compatibility):

- Added `jitter` (defaulting to `0.0` for storage, `0.2` in `defaultPubSubRetry`) and an injectable `isRetryable` predicate to `ExponentialRetry`.
- Since Pub/Sub's client already maps `GrpcError` status codes to `ServiceException`s with `Status(code: ...)`, `defaultIsRetryable` in `google_cloud_rpc` handles all mapped gRPC and HTTP retryable status codes directly without needing a `package:grpc` dependency. `google_cloud_pubsub` adds `isPubSubRetryable` (and `defaultPubSubRetry`) to cover raw `GrpcError`s in unit tests and Pub/Sub's backoff defaults (100 ms initial delay, 1.3× multiplier, ±20% jitter).
- Also fixed a subtle bug in `delaySequence`: `noRetriesAfter` is now computed eagerly when `delaySequence()` is called rather than inside the `sync*` generator body (which previously delayed deadline calculation until the first retry).

sigurdm:
> You didn't want to add any end-to-end tests?

You were right to push on this. I added four, and they immediately earned their keep.

**What's new** — `test/batching_e2e_test.dart` (tagged `firebase-emulator` and `google-cloud`):

1. A 25-message batch arrives as 25 distinct messages with 25 **distinct** message IDs. `Topic._onBatch` maps `messageIds[i] → batch[i]` positionally, and until now the fake was the only thing defining that contract.
2. 150 messages spanning several batches all arrive.
3. `Topic.close()` actually flushes what's still buffered — silent data loss on shutdown is the biggest risk batching introduces.
4. A batched ack really suppresses redelivery. `_onAckBatch` swallows errors as best-effort, so a malformed batch is invisible to unit tests but would cause infinite redelivery in production.

**And running the suite found a real bug** — in the stacked PR #339, not this one. Tearing down a streaming pull connection called `stream.drain()` whenever `StreamController.hasListener` was false. That's false in *two* situations: before the first listener subscribes, and after that listener has cancelled. In the second case `drain()` listens to an already-consumed single-subscription stream and throws a `StateError` **synchronously**, so the trailing `.catchError` never ran, the error escaped `cleanupCurrentConnection`, and reconnection stopped working entirely. 138 unit tests were green; the emulator test failed with `Bad state: Stream has already been listened to` followed by a 30 s hang. Fixed in #339 with a `DisposableStreamController` that tracks whether its stream was ever listened to, plus unit tests that fail against both mutants.

**Note on retry e2e coverage:** retry-on-`UNAVAILABLE` isn't covered here because the Pub/Sub emulator has no fault injection (unlike `storage-testbench`'s `return-503`).

pkgs/google_cloud_storage/lib/src/retry.dart
```diff
@@ -12,170 +12,4 @@
...
-
-/// The default retry strategy.
-///
-/// This strategy implements exponential backoff for [idempotent operations].
-///
-/// See [Retry strategy](https://docs.cloud.google.com/storage/docs/retry-strategy).
-///
-/// [idempotent operations]: https://docs.cloud.google.com/storage/docs/retry-strategy#idempotency-operations
-const defaultRetry = ExponentialRetry();
+export 'package:google_cloud_rpc/retry.dart';
```
brianquinlan, line 15: Shouldn't you move/delete retry_test.dart?

pkgs/google_cloud_pubsub/test/batching_e2e_test.dart
```diff
@@ -0,0 +1,182 @@
...
+library;
+
+import 'dart:convert';
+
+import 'package:google_cloud_pubsub/google_cloud_pubsub.dart';
+import 'package:test/test.dart';
+
+import 'test_utils.dart';
+
+void main() {
```
brianquinlan: Haven't we been combining the e2e tests and the mock tests in one test file for other tests?

brianquinlan (commented):
This change is 5000 LOC and hand to review. Is these an obvious way to split it up a bit?

@sigurdm do you have a good mental model of what this change is doing? I'm a bit worried that this PR is being driven by AI without us having a good understanding of what it is doing.

pkgs/google_cloud_storage/lib/src/retry.dart
sigurdm: Good catch — moved both `pkgs/google_cloud_storage/lib/src/retry.dart` and `pkgs/google_cloud_storage/test/retry_test.dart` to `generated/google_cloud_rpc` in #352.

pkgs/google_cloud_pubsub/test/batching_e2e_test.dart
sigurdm: Yes, done — I combined the mock and e2e tests into `test/publish_test.dart`, `test/acknowledge_test.dart`, `test/modify_ack_deadline_test.dart`, and `test/streaming_pull_test.dart` (with shared fakes in `test/test_utils.dart`) and deleted `batching_e2e_test.dart`, `topic_test.dart`, `subscription_batching_test.dart`, `lifecycle_test.dart`, and `wire_size_test.dart`.

sigurdm:
Fair point — a big chunk of the 5,000 LOC was redundant AI-generated test files (`lifecycle_test.dart`, `topic_test.dart`, `subscription_batching_test.dart`, `batching_e2e_test.dart`, `wire_size_test.dart`) that duplicated fake gRPC client boilerplate across every file instead of putting mock and e2e tests together in `publish_test.dart` / `acknowledge_test.dart` / `modify_ack_deadline_test.dart`.

I cleaned up those tests (cutting ~2,700 LOC) and split the change into a stack of 4 smaller PRs:

1. **#352** — Extract shared `RetryRunner` and `ExponentialRetry` into `package:google_cloud_rpc` (closes #346; split into two commits: one moving the files from `google_cloud_storage` and a second with the updates)
2. **#353** — Add `google_cloud_pubsub` `defaultRetry` and gRPC error mapping
3. **#354** — Add `Batcher` and `Topic` publish batching
4. **#292** — Add `Subscription` ack and `modifyAckDeadline` batching

Could we start by reviewing just **#352** (the extraction of the retry functionality)?

kevmoo:
https://github.com/kevmoo/analytica.dart - if you want some stuff to help dedupe bits, etc

PR #301, opened 2026-06-25
feat(storage): virtual filesystem in `listObjects`

brianquinlan:


pkgs/google_cloud_storage/lib/src/client.dart
```diff
@@ -862,6 +862,19 @@ final class Storage {
   /// A stream of objects contained in [bucket] in lexicographical order by
   /// name.
   ///
+  /// If [delimiter] is set, returns results in a directory-like mode, with
+  /// `'/'` being a common value for the delimiter. The result will only
+  /// include objects whose names do not contain `delimiter`, or whose names
+  /// only have instances of `delimiter` in their `prefix`. Note that common
+  /// prefixes (directories) returned by the API are not included in the
+  /// returned stream.
```
natebosch, line 870: Is this prose taken from somewhere else or original? I don't think I can fully follow it.

natebosch (approved):

pkgs/google_cloud_storage/lib/src/client.dart
brianquinlan: Yeah, I copied it from the API docs. I added an example.

PR #304, opened 2026-07-14
feat(storage): support ifMetagenerationNotMatch

### LINKED ISSUE #115: google_cloud_storage: Support `ifMetagenerationNotMatch`
If the parameter is passed to `package:googleapis` `BucketResource.patch` and the metageneration **does** match then the server returns a 304 with an empty body and `BucketResource.patch` tries to parse the empty body, failing with this exception:

```
  type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast

  package:googleapis/storage/v1.dart 1250:38          BucketsResource.patch
  ===== asynchronous gap ===========================
  package:google_cloud_storage/src/client.dart 25:12  _translateException
  ===== asynchronous gap ===========================
  package:google_cloud_storage/src/client.dart 103:5  Storage.patchBucket
  ===== asynchronous gap ===========================
  test/patch_bucket_test.dart 195:31                  main.<fn>.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 242:9    Declarer.test.<fn>.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 240:7    Declarer.test.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/invoker.dart 282:9     Invoker._waitForOutstandingCallbacks.<fn>
```

Amayyas:
Adds an `ifMetagenerationNotMatch` parameter to `Storage.patchBucket`,
`Storage.uploadObject` and `Storage.uploadObjectFromString`, resolving the three
TODOs that referenced this issue.

### The failure mode described in #115 no longer applies

The issue reports a `TypeError` during JSON deserialization, coming from
`package:googleapis`' `BucketsResource.patch`. That dependency is gone: both
request paths (`ServiceClient._makeRequest` and `uploadFile`) now reject any
non-2xx response before deserializing, so a 304 never reaches `jsonDecode`.

What a 304 *does* produce today is a bare `ServiceException` with
`statusCode: 304` and the message `"unknown error"`, which is why this PR still
adds a translation step.

### Behaviour

Verified against the storage-testbench: an unsatisfied `ifMetagenerationNotMatch`
returns **304 Not Modified with an empty body** — on `buckets.patch` and
`objects.insert` alike — rather than the **412** used by the `*Match`
preconditions. This PR maps that 304 to a new `NotModifiedException`.

`ServiceException` is a `final class`, so `NotModifiedException` cannot extend it
and is declared alongside `ChecksumValidationException` instead. If you would
rather have a generic `304 -> NotModifiedException` mapping in
`google_cloud_rpc` (i.e. in Librarian), I am happy to pursue that instead — it
would be reusable across APIs.

### Tests

* 5 unit tests (`MockClient`) covering the query parameter and the 304
  translation, including a check that a 412 is still reported as a
  `PreconditionFailedException`.
* 3 `storage-testbench` tests asserting the real server behaviour, and that the
  bucket/object is genuinely left unchanged.

`dart format`, `dart analyze`, `dart test` (272 passing) and
`dart test -P storage-testbench` (75 passing) are all clean.

Fixes #115


brianquinlan:
1.

Maybe the exception decoding logic should live in `package:google_cloud_rpc` around [here](https://github.com/googleapis/google-cloud-dart/blob/f1c2092b13a741d7e59ee743440c51527aa81e43/generated/google_cloud_rpc/lib/src/exceptions.dart#L61)

This seems consistent with the approach taken by other languages e.g. [python with `NotModified`(https://github.com/googleapis/google-cloud-python/blob/9eead955b654bec382106e08fbf1baddb250b297/packages/google-api-core/google/api_core/exceptions.py#L239)

2.

Could you move the `ifMetadataNotMatch` tests into the tests for the method that accepts that argument?

Amayyas:
Thanks — both done.

**1.** `304` is now mapped in `ServiceException._fromDecodedResponse` alongside the other status codes, so `NotModifiedException` is a real `ServiceException` subclass available to every API. The `_translateNotModified` wrapper and the storage-local class are gone. (My note in the PR description about `final class` was wrong — subclassing works fine from within the same library.)

One thing to flag: a 304 never has a body, so it was hitting the empty-body branch and surfacing as `NotModifiedException: unknown error`. `fromHttpResponse` now uses `'the resource was not modified'` instead. Happy to split that out if you'd rather keep this PR narrow.

**2.** Tests moved into `patch_bucket_test.dart`, `upload_object_test.dart` and `upload_object_from_string_test.dart`; the status mapping is covered in `google_cloud_rpc`'s `exceptions_test.dart`.

I left `google_cloud_rpc` at `0.5.3` (unreleased) and moved the storage entry to a new `0.6.4-wip` section since 0.6.3 is published — let me know if versioning is handled elsewhere.

generated/google_cloud_rpc/lib/src/exceptions.dart
```diff
@@ -231,6 +241,22 @@ final class ServiceException implements Exception {
   String toString() => '$_name: $message';
 }
 
+/// Exception thrown when the server returns a "304 Not Modified" response.
+///
+/// This indicates that the requested operation was not performed because a
```
brianquinlan: Could you remove these lines? They might only apply to `package:google_cloud_storage`.

pkgs/google_cloud_storage/test/patch_bucket_test.dart
```diff
@@ -916,6 +916,85 @@ void main() async {
       });
     });
 
+    group('storage-testbench', tags: ['storage-testbench'], () {
```
brianquinlan: Could you put this in the previous group instead of creating a new one?

pkgs/google_cloud_storage/test/patch_bucket_test.dart
```diff
@@ -916,6 +916,85 @@ void main() async {
...
+            BucketMetadataPatchBuilder()..labels = {'color': 'red'},
+            ifMetagenerationNotMatch: created.metageneration! + BigInt.one,
+          );
+
+          expect(actual.labels, containsPair('color', 'red'));
+        },
+      );
+    });
+
+    test('ifMetagenerationNotMatch is sent as a query parameter', () async {
```
brianquinlan: Please remove this test.

pkgs/google_cloud_storage/test/upload_object_from_string_test.dart
```diff
@@ -122,5 +124,27 @@ void main() {
 
       uploadObjectFromStringTest(() => storage, createBucketWithTearDown);
     });
+
+    test('ifMetagenerationNotMatch is sent as a query parameter', () async {
```
brianquinlan: Could you remove this test? I don't think that it is useful.

pkgs/google_cloud_storage/test/upload_object_test.dart
```diff
@@ -289,6 +289,71 @@ void main() async {
...
+          final content = await testbenchStorage.downloadObject(
+            bucketName,
+            'object1',
+          );
+          expect(utf8.decode(content), 'Hello World!');
+        },
+      );
+    });
+
+    test('ifMetagenerationNotMatch is sent as a query parameter', () async {
```
brianquinlan: Could you remove this test?

pkgs/google_cloud_storage/test/upload_object_test.dart
```diff
@@ -289,6 +289,71 @@ void main() async {
       });
     });
 
+    group('storage-testbench', tags: ['storage-testbench'], () {
```
brianquinlan: Could you add this to the previous group instead of creating a new one?

brianquinlan:
> Thanks — both done.
> 
> **1.** `304` is now mapped in `ServiceException._fromDecodedResponse` alongside the other status codes, so `NotModifiedException` is a real `ServiceException` subclass available to every API. The `_translateNotModified` wrapper and the storage-local class are gone. (My note in the PR description about `final class` was wrong — subclassing works fine from within the same library.)
> 
> One thing to flag: a 304 never has a body, so it was hitting the empty-body branch and surfacing as `NotModifiedException: unknown error`. `fromHttpResponse` now uses `'the resource was not modified'` instead. Happy to split that out if you'd rather keep this PR narrow.

No, what you did is great!

> 
> **2.** Tests moved into `patch_bucket_test.dart`, `upload_object_test.dart` and `upload_object_from_string_test.dart`; the status mapping is covered in `google_cloud_rpc`'s `exceptions_test.dart`.
> 
> I left `google_cloud_rpc` at `0.5.3` (unreleased) and moved the storage entry to a new `0.6.4-wip` section since 0.6.3 is published — let me know if versioning is handled elsewhere.

That's the correct way to do it, thanks!

The philosophy of testing in this package is to do integration tests against a GCP test project whenever possible.

brianquinlan (changes requested):

Amayyas:
Thanks, all addressed:

- **Integration tests** — moved the `patchBucket` and `uploadObject` cases into the existing `google-cloud` groups so they run against a real project, and dropped the MockClient unit tests (patch/upload/upload-from-string).
- **rpc exception doc** — removed the storage-specific sentence from `NotModifiedException`; it's just "304 Not Modified" now.

One judgement call: I left the single mapping test in `google_cloud_rpc`'s `exceptions_test.dart` (304 → `NotModifiedException`), since that's the generic layer where the decoding now lives and it doesn't need a backend. Happy to drop it too if you'd rather keep everything as integration tests.

`/gcbrun` when you get a chance — the integration tests only run there.

brianquinlan (approved):
Amazing contribution, thank you so much!

PR #309, opened 2026-07-17
fix(protobuf): reject cyclic collections when decoding Struct/Value

### LINKED ISSUE #285: _StructHelper.decode should deal with cyclic maps
For example:

```dart
final x = {}
x['x'] = x;
Value.from(x);
```

Probably this should throw.

Amayyas:
Decoding a cyclic Dart collection recursed until the stack overflowed:

```dart
final map = <String, Object?>{};
map['x'] = map;
Value.fromJson(map); // StackOverflowError
```

`Value`, `Struct` and `ListValue` decode into each other, so the decoding functions now thread the set of containers currently being decoded through the recursion, and throw a `FormatException` when one is re-entered.

### Notes

* **Lists have the same problem**, as do cycles that span both (a map holding a list that holds the map). Those are covered too, since the fix sits on the shared path.
* **`FormatException`** was chosen for consistency with the existing behaviour for other undecodable input — `Value.fromJson(DateTime.now())` already throws it. Happy to switch to `ArgumentError` if you'd rather treat it as a caller-side mistake.
* **Repeated-but-acyclic input still decodes.** Entries leave the set once decoded, so only ancestors are rejected — e.g. the same map referenced by two keys is fine. Covered by tests.
* **Only decoding is affected.** `Struct.fields` is `final`, so a cyclic object graph cannot be constructed in the first place and encoding cannot recurse.

`dart format`, `dart analyze`, `dart test` (180 passing) and the protojson conformance suite (211 passing) are all clean.

Fixes #285


generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -549,21 +530,76 @@ class _StructHelper {
...
       value.values.map((v) => v.toJson()).toList();
 
-  static ListValue decode(Object? value) {
-    final values = (value as List).map(Value.fromJson).toList();
-    return ListValue(values: values);
+  static ListValue decode(Object? value) =>
+      _decodeListValue(value, Set.identity());
+}
+
+// `Value`, `Struct` and `ListValue` decode into each other, so the decoding
```
brianquinlan, line 544: This should be dartdoc i.e. start with `///`.

Maybe something simpler:

```dart
/// Decodes [json] into a [Value].
///
/// Tracks previously seen contains in [parents] to avoid infinite recursion when
/// decoding recursive data structures.
```

generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -549,21 +530,76 @@ class _StructHelper {
...
+      return Value(listValue: _decodeListValue(l, parents));
+    case Map<dynamic, dynamic> m:
+      return Value(structValue: _decodeStruct(m, parents));
+    default:
+      throw FormatException('Invalid Value: $json');
+  }
+}
+
+/// Decodes [json] into a [Struct].
+Struct _decodeStruct(Object? json, Set<Object> parents) {
```
brianquinlan, line 575:
```suggestion
Struct _decodeStruct(Map<String, dynamic> json, Set<Object> parents) {
```

generated/google_cloud_protobuf/lib/src/protobuf.p.dart
```diff
@@ -549,21 +530,76 @@ class _StructHelper {
...
+        (key, value) => MapEntry(key, _decodeValue(value, parents)),
+      ),
+    );
+  } finally {
+    parents.remove(map);
+  }
+}
+
+/// Decodes [json] into a [ListValue].
+ListValue _decodeListValue(Object? json, Set<Object> parents) {
```
brianquinlan, line 592:
```suggestion
ListValue _decodeListValue(List<dynamic> json, Set<Object> parents) {
```

brianquinlan:
@Amayyas are you still working on this?

PR #321, opened 2026-08-18
feat: add setup and test agentic skills 

brianquinlan:
Adds the new file `generated/RELEASING.md`, which describes the first steps in the generated-code release process. I will add more steps to that process as I complete them.

Restructures `generated/README.md` to avoid duplication with `generated/RELEASING.md`.

Modifies some test file to avoid duplicate imports now that every `google_cloud_*` library exports `package:google_cloud_rpc/exceptions.dart`.

The remaining changes are done through automation by completing steps 1-4 from `generated/RELEASING.md`.

### DIFF
--- a/.github/workflows/dart_checks.yaml
+++ b/.github/workflows/dart_checks.yaml
@@ -27,7 +27,7 @@ on:
 env:
   DART_VERSION: 3.9.2
   GOOGLE_CLOUD_PROJECT: test-project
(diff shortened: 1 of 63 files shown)


natebosch (approved):
Should we be adding some of these files to `.gitattributes` to mark them as `linguist-generated`?

brianquinlan:
> Should we be adding some of these files to `.gitattributes` to mark them as `linguist-generated`?

Yes, I'll do that in a follow-up PR.

PR #329, opened 2026-08-25
feat: support client attribution header in ServiceClient and send gccl in Storage

kevmoo:
* Add optional `apiClientHeader` to `ServiceClient` in `package:google_cloud_rpc`.
* Send `gccl/<version>` in `package:google_cloud_storage` requests.
* Add unit tests verifying header emission in both packages.

### DIFF
--- a/generated/google_cloud_rpc/CHANGELOG.md
+++ b/generated/google_cloud_rpc/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.5-wip
+
+- feat: add optional `apiClientHeader` parameter to `ServiceClient`.
+
 ## 0.5.4
 
 - feat: add setup and test agentic skills  (#321)
--- a/generated/google_cloud_rpc/lib/service_client.dart
+++ b/generated/google_cloud_rpc/lib/service_client.dart
@@ -30,7 +30,7 @@ const String _clientKey = 'x-goog-api-client';
 
 // ignore: prefer_const_declarations
 final String _clientName =
-    'gl-dart/$clientDartVersion gax/$packageVersion rest/$packageVersion gapic/$packageVersion';
+    'gl-dart/$clientDartVersion gax/$packageVersion rest/$packageVersion';
 
 const String _contentTypeKey = 'content-type';
 const String _typeJson = 'application/json';
--- a/generated/google_cloud_rpc/lib/service_client.dart
+++ b/generated/google_cloud_rpc/lib/service_client.dart
@@ -31,8 +31,6 @@ const String _clientKey = 'x-goog-api-client';
 String get _baseClientName =>
     'gl-dart/$clientDartVersion gax/$packageVersion rest/$packageVersion';
 
-String get _clientName => '$_baseClientName gapic/$packageVersion';
-
 const String _contentTypeKey = 'content-type';
 const String _typeJson = 'application/json';
 
@@ -40,28 +38,60 @@ const String _typeJson = 'application/json';
 class ServiceClient {
   final http.Client client;
 
-  /// An optional API client header value (e.g. `gccl/0.6.4` or `gapic/0.1.0`)
-  /// to append to `x-goog-api-client`.
-  final String? apiClientHeader;
+  /// The version of the GAPIC-generated client library (e.g. `0.1.0`),
+  /// which will be formatted as `gapic/<version>` in the `x-goog-api-client`
+  /// header.
+  ///
+  /// Set this for client libraries that were generated completely by
+  /// automation.
+  final String? gapicVersion;
+
+  /// The version of the hand-written client library (e.g. `0.6.4`),
+  /// which will be formatted as `gccl/<version>` in the `x-goog-api-client`
+  /// header.
+  ///
+  /// Set this if the client library includes hand-written components.
+  final String? gcclVersion;
 
   /// Creates a `ServiceClient` using [client] for transport.
+  ///
+  /// If [gapicVersion] is set, `gapic/<version>` is included in the
+  /// `x-goog-api-client` header.
+  ///
+  /// If [gcclVersion] is set, `gccl/<version>` is included in the
+  /// `x-goog-api-client` header.
+  ///
+  /// Both [gapicVersion] and [gcclVersion] should be set if the library
+  /// includes both automatically-generated and hand-written components.
+  /// For example, `package:google_cloud_firestore` (hand-written) wraps
+  /// `package:google_cloud_firestore_v1` (generated).
   ///
   /// The provided [http.Client] must be configured to provide whatever
   /// authentication is required by the API being accessed. You can do that
   /// using
   /// [`package:googleapis_auth`](https://pub.dev/packages/googleapis_auth).
-  ServiceClient({required this.client, this.apiClientHeader});
+  ServiceClient({required this.client, this.gapicVersion, this.gcclVersion});
 
   /// The `x-goog-api-client` header value sent with every request.
   String get clientHeader {
-    if (apiClientHeader != null && apiClientHeader!.trim().isNotEmpty) {
-      final custom = apiClientHeader!.trim();
-      if (custom.startsWith('gccl/') || custom.startsWith('gapic/')) {
-        return '$_baseClientName $custom';
-      }
-      return '$_clientName $custom';
+    final buffer = StringBuffer(_baseClientName);
+    final gapic = gapicVersion?.trim();
+    final hasGapic = gapic != null && gapic.isNotEmpty;
+    final gccl = gcclVersion?.trim();
+    final hasGccl = gccl != null && gccl.isNotEmpty;
+
+    if (hasGapic) {
+      buffer.write(' gapic/$gapic');
+    } else if (!hasGccl) {
+      // Default fallback to packageVersion for un-updated generated clients.
+      buffer.write(' gapic/$packageVersion');
     }
-    return _clientName;
+
+    if (hasGccl) {
+      buffer.write(' gccl/$gccl');
+    }
+
+    return buffer.toString();
   }
 
   Future<Map<String, dynamic>> get(Uri url) => _makeRequest(url, 'GET');
--- a/generated/google_cloud_rpc/lib/service_client.dart
+++ b/generated/google_cloud_rpc/lib/service_client.dart
@@ -30,7 +30,7 @@ const String _clientKey = 'x-goog-api-client';
 
 // ignore: prefer_const_declarations
 final String _clientName =
-    'gl-dart/$clientDartVersion gax/$gaxVersion rest/$gaxVersion gapic/$gaxVersion';
+    'gl-dart/$clientDartVersion gax/$packageVersion rest/$packageVersion gapic/$packageVersion';
 
 const String _contentTypeKey = 'content-type';
 const String _typeJson = 'application/json';
@@ -50,6 +50,12 @@ class ServiceClient {
   /// using
   /// [`package:googleapis_auth`](https://pub.dev/packages/googleapis_auth).
   ServiceClient({required this.client, this.apiClientHeader});
+
+  /// The `x-goog-api-client` header value sent with every request.
+  String get clientHeader =>
+      apiClientHeader != null && apiClientHeader!.trim().isNotEmpty
+      ? '$_clientName ${apiClientHeader!.trim()}'
+      : _clientName;
 
   Future<Map<String, dynamic>> get(Uri url) => _makeRequest(url, 'GET');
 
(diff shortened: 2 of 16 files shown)


generated/google_cloud_rpc/test/client_test.dart
```diff
@@ -61,6 +61,25 @@ void main() {
...
+          }),
+          apiClientHeader: 'gccl/0.6.4-wip',
+        );
+
+        await customService.get(sampleUrl);
+
+        expect(customRequest.headers, {
+          'x-goog-api-client': matches(
+            r'^gl-dart/(3\.\d+\.\d+)|0 gax/0.2.0 rest/0.2.0 gapic/0.2.0 gccl/0.6.4-wip$',
+          ),
```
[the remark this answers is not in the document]
kevmoo: Fixed regex assertion to group the alternation and escape literal dots.

pkgs/google_cloud_storage/test/storage_test.dart
```diff
@@ -127,13 +128,21 @@ void main() async {
...
-      await expectLater(
-        storage.listBuckets(),
-        emitsInOrder([emitsError(isA<StateError>()), emitsDone]),
+      await storage.listBuckets().drain<Object?>();
+
+      expect(
+        actualRequest.headers['x-goog-api-client'],
+        contains('gccl/0.6.4-wip'),
       );
     });
```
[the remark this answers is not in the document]
kevmoo: Restored the listBuckets StateError test in storage_test.dart.

generated/google_cloud_rpc/lib/service_client.dart
```diff
@@ -39,13 +39,23 @@ const String _typeJson = 'application/json';
 class ServiceClient {
   final http.Client client;
 
+  /// An optional API client header value (e.g. `gccl/0.6.4`) to append to
+  /// `x-goog-api-client`.
+  final String? apiClientHeader;
```
brianquinlan: I'm sure sure how this is supposed to work . For Rust, this is only set for some APIs (https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-rust+gccl&type=code), including storage. 

Docs say:

gccl: google-cloud (Veneer) Client Library Version
gapic: Version of the GAPIC (Veneer Toolkit) Generated Client Library

If I'm interpreting [this Rust code](https://github.com/googleapis/google-cloud-rust/blob/384be4abcfc811d9ce9134ec2bd8ae72eaf963ff/src/gax-internal/src/api_header.rs#L51) correctly, then we should:

`gl-dart/<dart version> gax/<google_cloud_rpc version> rest/<google_cloud_rpc version> gapic/<version of actual package>`

If the package is hand-written, then we also set `gccl` to the package's version.

Is that your interpretation too?

I'm looking at [go/cloud-api-headers](http://go/cloud-api-headers)

brianquinlan: Would it make sense to make the library version available in the generated client, i.e., https://github.com/googleapis/librarian/pull/7403

And then `ServiceClient` could take a parameter to collect this version? The Rust client has `library_type` and `version` parameters (https://github.com/googleapis/google-cloud-rust/blob/384be4abcfc811d9ce9134ec2bd8ae72eaf963ff/src/gax-internal/src/api_header.rs#L21).

kevmoo: Yes, that matches the AIP-210 / go/cloud-api-headers specification! Base `google_cloud_rpc` now emits `gl-dart/<dart> gax/<rpc> rest/<rpc>`, hand-written veneer packages (like Storage) pass `gccl/<version>`, and generated packages can pass `gapic/<version>` once https://github.com/googleapis/librarian/pull/7403 lands. I have updated `_clientName` in `ServiceClient` accordingly.

pkgs/google_cloud_storage/lib/src/client.dart
```diff
@@ -95,7 +96,10 @@ final class Storage {
   };
 
   FutureOr<ServiceClient> get _serviceClient async =>
-      _cachedServiceClient ??= ServiceClient(client: await _httpClient);
+      _cachedServiceClient ??= ServiceClient(
+        client: await _httpClient,
+        apiClientHeader: 'gccl/$packageVersion',
+      );
```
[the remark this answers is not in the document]
kevmoo: Refactored _serviceClient getter to return the cached ServiceClient synchronously if populated, avoiding Future wrapping and microtask overhead.

pkgs/google_cloud_storage/lib/src/file_upload.dart
```diff
@@ -89,6 +90,9 @@ Future<ObjectMetadata> uploadFile(
   final request = http.Request('POST', url);
   request.headers['Content-Type'] = 'multipart/related; boundary=$boundary';
   request.headers['Content-Length'] = bodyBytes.length.toString();
+  if (headers != null) {
+    request.headers.addAll(headers);
+  }
```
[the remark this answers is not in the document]
kevmoo: Added custom headers before setting Content-Type and Content-Length so user headers cannot overwrite multipart protocol headers.

pkgs/google_cloud_storage/lib/src/resumeable_upload.dart
```diff
@@ -178,11 +180,11 @@ class ResumableUploadSink implements StreamSink<List<int>> {
...
-        final headers = {'Content-Range': contentRange};
-        if (hashHeader != null) {
-          assert(isLast);
-          headers['x-goog-hash'] = hashHeader;
-        }
+        final headers = {
+          'Content-Range': contentRange,
+          'x-goog-hash': ?hashHeader,
+          ...?resolvedHeaders,
+        };
```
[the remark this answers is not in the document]
kevmoo: Restored assert(isLast) to verify hash header is only passed on the final chunk.

generated/google_cloud_rpc/lib/src/versions.dart
```diff
@@ -16,10 +16,9 @@ export 'web.dart' if (dart.library.io) 'vm.dart' show clientDartVersion;
...
-// * lib/src/versions.dart
-/// The format is either `major.minor.patch` or the special value `0`, which
-/// indicates that the version is unknown.
-const gaxVersion = '0.2.0';
+// TODO: consider using package:build_version or similar to automate
+// syncing with pubspec.yaml.
+const packageVersion = '0.5.5-wip';
+
+@Deprecated('Use packageVersion instead')
+const gaxVersion = packageVersion;
```
brianquinlan: Is this exported? Can we just remove it?

generated/google_cloud_rpc/lib/service_client.dart
brianquinlan: Is this the best representation?

[Rust](https://github.com/googleapis/google-cloud-rust/blob/384be4abcfc811d9ce9134ec2bd8ae72eaf963ff/src/gax-internal/src/api_header.rs#L22) lets you set the API type (e.g. "gapic" or "gccl") and the library version.

[Python](https://github.com/googleapis/google-cloud-python/blob/082a99a2c4a3e8d5df28eaeab9b2c710dd4296d5/packages/google-api-core/google/api_core/client_info.py#L46) lets the client pass in whatever values they want.

generated/google_cloud_rpc/lib/service_client.dart
```diff
@@ -39,13 +40,29 @@ const String _typeJson = 'application/json';
...
+
   /// Creates a `ServiceClient` using [client] for transport.
   ///
   /// The provided [http.Client] must be configured to provide whatever
   /// authentication is required by the API being accessed. You can do that
   /// using
   /// [`package:googleapis_auth`](https://pub.dev/packages/googleapis_auth).
-  ServiceClient({required this.client});
+  ServiceClient({required this.client, this.apiClientHeader});
```
brianquinlan:
```suggestion
  /// If [gapicVersion] is set then `gapic/<version>` will be included in the `x-goog-api-client` header.
  /// [gapicVersion] should be set for client libraries that were generated completely by automation.
  ///
  /// If [gcclVersion] is set then `gccl/<version>` will be included in the `x-google-api-client` header.
  /// [gcclVersion] should be set if the client library includes hand-written components.
  ///
  /// Both [gapicVersion] and [gcclVersion] should be set if the client client includes both automatically-generated and hand-written components. For example, `package:google_cloud_firestore` (hand-written) depends on `package:google_cloud_firestore_v1` (generated). So it should set `gapic/<google_cloud_firestore_v1 version> gccl/<google_cloud_firestore version`.
  ServiceClient({required this.client, this.gapicVersion, this.gcclVersion});
```

To accomplish this, I can change the constructor for the generated services to include a `gcclVersion` parameter, which we can use from `Firestore` (https://github.com/firebase/firebase-admin-dart/blob/27a3be7d2afc343ad942a30a73c5399deb519ccc/packages/google_cloud_firestore/lib/src/firestore_http_client.dart#L245).

I can also change the construction of `ServiceClient` from the generated clients to pass `gapicVersion`. The constructor signature of the Python clients is to take a `client_info`, with this definition:

```
Optional[google.api_core.gapic_v1.client_info.ClientInfo]
The client info used to send a user-agent string along with API requests. If None, then default info will be used. Generally, you only need to set this if you're developing your own library or partner tool.'

google.api_core.client_info.ClientInfo(python_version='3.10.19', grpc_version='1.83.0', api_core_version='2.34.0', gapic_version=None, client_library_version=None, user_agent=None, rest_version=None, protobuf_runtime_version=None)
```

That seems a bit heavy.

WDYT? Any gaps in that idea?

generated/google_cloud_rpc/lib/src/versions.dart
kevmoo: Confirmed it's internal to lib/src/versions.dart and not exported. Removed!

generated/google_cloud_rpc/lib/service_client.dart
kevmoo: Adopted! Updated `ServiceClient` constructor to take `this.gapicVersion` and `this.gcclVersion`. This cleanly models the layered Firestore (`gapic` from generated client + `gccl` from veneer) and hand-written Storage (`gccl` only) cases without string parsing or ClientInfo boilerplate.

generated/google_cloud_rpc/lib/src/versions.dart
```diff
@@ -16,10 +16,6 @@ export 'web.dart' if (dart.library.io) 'vm.dart' show clientDartVersion;
...
-// Keep these versions in sync:
-// * CHANGELOG.yaml
-// * pubspec.yaml
-// * lib/src/versions.dart
-/// The format is either `major.minor.patch` or the special value `0`, which
-/// indicates that the version is unknown.
-const gaxVersion = '0.2.0';
+// TODO: consider using package:build_version or similar to automate
+// syncing with pubspec.yaml.
+const packageVersion = '0.5.5-wip';
```
brianquinlan, line 21: We don't use -wip for generated packages. But I think that I'll end up nuking this file the next time that the generator runs anway.

generated/google_cloud_rpc/lib/service_client.dart
```diff
@@ -39,13 +38,61 @@ const String _typeJson = 'application/json';
...
+  /// If [gapicVersion] is set, `gapic/<version>` is included in the
+  /// `x-goog-api-client` header.
+  ///
+  /// If [gcclVersion] is set, `gccl/<version>` is included in the
+  /// `x-goog-api-client` header.
+  ///
+  /// Both [gapicVersion] and [gcclVersion] should be set if the library
+  /// includes both automatically-generated and hand-written components.
+  /// For example, `package:google_cloud_firestore` (hand-written) wraps
+  /// `package:google_cloud_firestore_v1` (generated).
```
brianquinlan, line 67: Can you clarify that `[gapicVersion]` would be for `..firestore_v1` and `[gcclVersion]` would be for `...firestore`.

generated/google_cloud_rpc/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.5-wip
+
+- feat: add optional `apiClientHeader` parameter to `ServiceClient`.
```
brianquinlan, line 5: This is generated code, you shouldn't write to it (but your commit will be added automatically).

generated/google_cloud_rpc/pubspec.yaml
```diff
@@ -16,7 +16,7 @@
 
 name: google_cloud_rpc
 description: The Google Cloud client library for the Google RPC Types.
-version: 0.5.4
+version: 0.5.5-wip
```
brianquinlan, line 19: You can't change this, it will get undone after generation.

pkgs/google_cloud_storage/pubspec.yaml
```diff
@@ -32,7 +32,7 @@ dependencies:
   ffi: ^2.2.0
   google_cloud: '>=0.3.0 <0.6.0'
   google_cloud_protobuf: ^0.5.0
-  google_cloud_rpc: ^0.5.0
+  google_cloud_rpc: ^0.5.5-wip
```
brianquinlan, line 35: If you wanted to do this cleanly, you'd probably need multiple PRs. The cleanest way is probably:
1. we change `package:google_cloud_rpc`
2. we change the generator to use inject the versions
3. we publish all of the generated packages
4. we update `package:google_cloud_storage`

If you reduce this PR to just (1) then I promised to do 2-4.

Issue #322, opened 2026-08-19
feat(auth): add package:google_cloud_auth for server-side Google ID token & OIDC verification


kevmoo:
Add a new package, `package:google_cloud_auth`, to `google-cloud-dart` under `pkgs/google_cloud_auth/`.

This package provides canonical, pure-Dart server-side authentication primitives for Google Cloud and Dart backends, enabling fast in-memory verification of inbound Google ID tokens (RS256) and OpenID Connect (OIDC) tokens against public JSON Web Key Sets (JWKS).

## Motivation & Ecosystem Context

In other Google Cloud language SDKs (Node.js `google-auth-library`, Python `google-auth`, Go `cloud.google.com/go/auth`, Java `google-auth-library-java`), developers have official libraries to verify inbound Google ID tokens locally against Google's public JWKS.

In Dart, this capability has been missing from official packages, forcing major downstream Dart systems into expensive workarounds:
- **`pub.dev` Login:** Makes remote HTTP round-trips to `oauth2.googleapis.com/tokeninfo` (a debug endpoint) on every login request.
- **`pub.dev` Automated Publishing:** Spawns a standalone Go child process (`pkg/signature_verifier`) and uses temporary disk files to verify RSA-SHA256 signatures on GitHub Actions and Google Cloud WIF OIDC tokens.
- **Firebase Dart SDKs:** Bundle duplicate JWKS caching engines relying on third-party crypto packages (`dart_jsonwebtoken` / `pointycastle`).

`package:google_cloud_auth` consolidates these use cases into a single, officially maintained Google package with zero native C/C++ compilation requirements.

## Scope & Capabilities

1. **Pure-Dart RS256 Verification Engine:**
   - Inlined `BigInt.modPow` + `package:crypto` SHA-256 (`EMSA-PKCS1-v1_5`).
   - Zero native build dependencies (no CMake or C/C++ compiler required).
   - Validated against 100% of NIST test vectors.
2. **JWS Wire Parser & Base64URL:**
   - Strict RFC 7515/7519 envelope splitting and unpadded base64url decoding.
3. **In-Memory JWKS Cache:**
   - Honors HTTP `Cache-Control: max-age` and `Age` headers.
   - Handles key rotation automatically via single cache-busting re-fetch on unknown `kid`.
4. **Generic `IdTokenVerifier`:**
   - Verifies arbitrary OIDC tokens (GitHub Actions, Firebase, Auth0, etc.).
[...]

PR #337, opened 2026-09-04
feat(auth): add compute engine credentials

brianquinlan:
Part of https://github.com/googleapis/google-cloud-dart/issues/322

The important addition is a new `ComputeEngineCredentials` class, which allows key signing using the `serviceAccounts.signBlob` RPC.

The architecture is heavily copied from Python and Java.

About 25 LOC of the metadata server code is duplicated with `package:google_cloud` but it didn't seem worth it to add the extra required metadata features to `package:google_cloud` and then add a dependency. I might change my mind about that later.


pkgs/google_cloud_auth/lib/src/compute_engine_credentials.dart
```diff
@@ -0,0 +1,368 @@
...
+  final String metadataHost;
+
+  final http.Client _client;
+  final bool _ownsClient;
+
+  String? _cachedAccessToken;
+  DateTime? _accessTokenExpiry;
+
+  /// The active request to fetch a new token. This is used to prevent
+  /// multiple concurrent requests from fetching new tokens at the same time.
```
natebosch: [nit] https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

```suggestion
  /// The active request to fetch a new token.
  ///
  /// This is used to prevent multiple concurrent requests from fetching new
  /// tokens at the same time.
```

pkgs/google_cloud_auth/lib/src/compute_engine_credentials.dart
```diff
@@ -0,0 +1,368 @@
...
+  ComputeEngineCredentials._({
+    required this.clientEmail,
+    required this.universeDomain,
+    required this.metadataHost,
+    required http.Client client,
+    required bool ownsClient,
+  }) : _client = client,
+       _ownsClient = ownsClient;
+
+  /// Returns a non-expired current OAuth2 access token for the service account.
```
natebosch:
```suggestion
  /// A non-expired current OAuth2 access token for the service account.
```

Avoid starting docs with "Returns".

pkgs/google_cloud_auth/lib/src/compute_engine_credentials.dart
```diff
@@ -0,0 +1,368 @@
...
+  ///
+  /// If the cached token is expired or if [forceRefresh] is `true`, then a new
+  /// token is fetched.
+  ///
+  /// It is safe to call this method concurrently.
+  ///
+  /// Throws [CredentialException] on failure.
+  Future<String> getAccessToken({bool forceRefresh = false}) async {
```
natebosch:
```suggestion
  Future<String> accessToken({bool forceRefresh = false}) async {
```
https://dart.dev/effective-dart/design#avoid-starting-a-function-or-method-name-with-get

pkgs/google_cloud_auth/test/compute_engine_credentials_test.dart
```diff
@@ -0,0 +1,759 @@
...
+          expect(request.headers['metadata-flavor'], 'Google');
+          final path = request.url.path;
+          if (path ==
+              '/computeMetadata/v1/instance/service-accounts/default/email') {
+            return http.Response('sa@test.iam.gserviceaccount.com', 200);
+          }
+          if (path == '/computeMetadata/v1/universe/universe-domain') {
+            return http.Response('custom.domain.com', 200);
+          }
+          return http.Response('Not found', 404);
```
natebosch:
```suggestion
          return switch (request.url.path) {
            '/computeMetadata/v1/instance/service-accounts/default/email' =>
              http.Response('sa@test.iam.gserviceaccount.com', 200),
            '/computeMetadata/v1/universe/universe-domain' => http.Response(
              'custom.domain.com',
              200,
            ),
            _ => http.Response('Not found', 404),
          };
```

I think if we're on a new enough language version for do shorthands this could also be

```dart
          return switch (request.url.path) {
            '/computeMetadata/v1/instance/service-accounts/default/email' =>
              .new('sa@test.iam.gserviceaccount.com', 200),
            '/computeMetadata/v1/universe/universe-domain' => .new(
              'custom.domain.com',
              200,
            ),
            _ => .new('Not found', 404),
          };
```

But there may be mixed opinions on going that far.

pkgs/google_cloud_auth/test/compute_engine_credentials_test.dart
```diff
@@ -0,0 +1,759 @@
...
+        final mockClient = MockClient((request) async {
+          final path = request.url.path;
+          if (path ==
+              '/computeMetadata/v1/instance/service-accounts/default/email') {
+            return http.Response('sa@test.iam.gserviceaccount.com', 200);
+          }
+          if (path == '/computeMetadata/v1/universe/universe-domain') {
+            return http.Response('Internal error', 500);
+          }
+          return http.Response('Not found', 404);
```
natebosch: [optional] consider refactoring to a switch expression

brianquinlan:
FYI @kevmoo

pkgs/google_cloud_auth/lib/src/compute_engine_credentials.dart
brianquinlan: Yeah, sorry, too much copying from Java

brianquinlan: Done

pkgs/google_cloud_auth/test/compute_engine_credentials_test.dart
brianquinlan: Done

pkgs/google_cloud_auth/test/compute_engine_credentials_test.dart
brianquinlan: Done.

natebosch (approved):

PR #342, opened 2026-09-08
feat(tasks_v2): initial support

brianquinlan:
Most of the code is generated.

Interesting changes:
- added an example
- added a simple change
- modified the terraform configuration to grant the test runner access to `cloudtasks.googleapis.com`


generated/google_cloud_tasks_v2/test/tasks_test.dart
```diff
@@ -0,0 +1,81 @@
...
+    late CloudTasks tasksService;
+    late String queueName;
+
+    setUp(() async {
+      final client = await auth.clientViaApplicationDefaultCredentials(
+        scopes: ['https://www.googleapis.com/auth/cloud-platform'],
+      );
+      tasksService = CloudTasks(client: client);
+      final parent = 'projects/$projectId/locations/$location';
+      final queueId = 'queue-${Random().nextInt(999999999)}';
```
natebosch, line 42: Are we unable to write tests with deterministic behavior?

I guess this is evidence in support of the need for https://github.com/dart-lang/test/issues/1572

natebosch (approved):

generated/google_cloud_tasks_v2/test/tasks_test.dart
brianquinlan: In my mind, this is no difference than creating a temp directory in a test with a random name - the test isn't about the directory and the name of the directory shouldn't be material.

PR #344, opened 2026-09-09
feat(auth): add a mechanism to get default creds

brianquinlan:
Part of https://github.com/googleapis/google-cloud-dart/issues/357

Creates a new abstract class `GoogleCredentials` and a new function `defaultCredentials`, which will return the currently configured default credentials. The implementation of the `defaultCredentials` function was largely copied from the [Python](https://github.com/googleapis/google-auth-library-python/blob/2ea24b03436765fa3cf279ce148482ff6332136b/google/auth/_default.py#L597) and [Java](https://github.com/googleapis/google-cloud-java/blob/main/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java) implementations.

Currently only `ComputeEngineCredentials` and `ServiceAccountCredentials` are supported.


kevmoo:
Is this deplicating bits from `googleapis_auth`? That's fine...we should just...note it. And ponder if/how we deprecate `googleapis_auth` ...even parts of it.

kevmoo:
I filed #357 to track the follow-up work for `AuthorizedUserCredentials`, `accessToken()`, and `AuthClient` so we can keep this PR focused on `findDefaultCredentials()` and `url_signer_test.dart`.

Two inline design questions for this PR related to #357:

1. **`GoogleCredentials` base contract (`accessToken()` vs. `ServiceAccountSigner` downcast)**:
   Right now `GoogleCredentials` only defines `universeDomain` and `close()`, while blob signing lives on `ServiceAccountSigner` (`clientEmail` + `sign()`), which forces `url_signer_test.dart` to downcast `(await findDefaultCredentials()) as ServiceAccountSigner`. Since every ADC credential type (`ServiceAccountCredentials`, `ComputeEngineCredentials`, and upcoming `AuthorizedUserCredentials` / `ExternalAccountCredentials`) can provide OAuth2 bearer tokens (`Authorization: Bearer <token>`), should `GoogleCredentials` define the token-provision contract (e.g., `Future<AccessToken> accessToken({bool forceRefresh = false})`) while keeping `ServiceAccountSigner` as the opt-in capability interface for credentials that support blob signing?

2. **`~/.config/gcloud/application_default_credentials.json` (`authorized_user`) behavior before `AuthorizedUserCredentials` lands**:
   In `findDefaultCredentials()` (lines 66–72), Step 2 checks `~/.config/gcloud/application_default_credentials.json`. Because `gcloud auth application-default login` always writes `"type": "authorized_user"` to that file, `_loadCredentialsFile` will currently throw `CredentialException("The credential at '...' has type 'authorized_user', which is not recognized.")` on any developer workstation or VM that has a `gcloud` ADC file, preventing fallthrough to `ComputeEngineCredentials`.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
```diff
@@ -0,0 +1,164 @@
...
+import 'package:path/path.dart' as p;
+
+import 'compute_engine_credentials.dart';
+import 'credential_exception.dart';
+import 'google_credentials.dart';
+import 'service_account_credentials.dart';
+
+// Design based on:
+// - https://github.com/googleapis/google-cloud-java/blob/main/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+// - https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_default.py
```
natebosch, line 29: [optional] Consider moving to a library level doc comment.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
```diff
@@ -0,0 +1,164 @@
...
+
+/// [defaultCredentials] with some extra parameters for testing.
+@internal
+Future<GoogleCredentials> internalDefaultCredentials({
+  http.Client? client,
+  @visibleForTesting String? Function(String name)? getEnvironmentVariable,
+  @visibleForTesting String? wellKnownFilePath,
+  @visibleForTesting bool? isWindows,
+}) async {
+  final getEnv = getEnvironmentVariable ?? (name) => Platform.environment[name];
```
natebosch, line 47: [optional] If you make the default implementation a top level private function we should be able to use it as a default argument value.

```dart
String? _defaultReadEnvironment(String name) => Platform.environment[name];

Future<GoogleCredentials> internalDefaultCredentials({
  http.Client? client,
  @visibleForTesting String? Function(String name) readEnvironment = _defaultReadEnvironment,
  @visibleForTesting String? wellKnownFilePath,
  @visibleForTesting bool? isWindows,
}) async {
```

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
```diff
@@ -0,0 +1,164 @@
...
+///
+/// Throws a [CredentialException] if no credentials could be found or loaded.
+Future<GoogleCredentials> defaultCredentials({http.Client? client}) =>
+    internalDefaultCredentials(client: client);
+
+/// [defaultCredentials] with some extra parameters for testing.
+@internal
+Future<GoogleCredentials> internalDefaultCredentials({
+  http.Client? client,
+  @visibleForTesting String? Function(String name)? getEnvironmentVariable,
```
natebosch, line 43: [nit] How about `readEnvironment` in the spirit of https://dart.dev/effective-dart/design#avoid-starting-a-function-or-method-name-with-get

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  throw CredentialException(
+    "The credential at '${file.path}' has type '$type', "
+    'which is not recognized.',
+  );
+}
+
+@internal
+String? getWellKnownCredentialsPath(
```
natebosch, line 145: [nit] Drop leading "get". Also I think this can be private? -`String _welKnownCredentialsPath(`

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
```diff
@@ -0,0 +1,164 @@
...
+  String? Function(String name) getEnv,
+  bool isWindows,
```
natebosch, line 147: Do we need to be able to fake the platform for testing, or can we get away with having CI run on both windows and other OSes?

pkgs/google_cloud_auth/lib/src/google_credentials.dart
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Design based on:
+// - https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+// - https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py
+
+/// Base class for credentials used to authenticate with Google APIs and
+/// services.
+abstract class GoogleCredentials {
```
natebosch, line 21: Should we use the `base` keyword?

If we do that we can also avoid the OOP related language in the docs because it's already documented by the signature.

```suggestion
/// Credentials used to authenticate with Google APIs and services.
abstract base class GoogleCredentials {
```

pkgs/google_cloud_auth/test/test_utils.dart
```diff
@@ -0,0 +1,79 @@
...
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:webcrypto/webcrypto.dart';
+
+final canUseWebCrypto = () {
+  // package:webcrypto is always available in the browser.
+  // Dart 3.13 or later is required when run using the vm.
+  if (!const bool.fromEnvironment('dart.library.io')) return true;
```
natebosch, line 23: There is an unconditional import of `dart:io` above - this library shouldn't be available unless this would be true...

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: Do you mean as dartdoc? I was intending this as a sign post to help other implementers of this package not something that users should be aware of.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: Done.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: Done. There is a note in AGENTS.md that functions shouldn't start with "get". I'm surprised that gemini code review didn't look at that and flag it during review.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
natebosch: > I was intending this as a sign post to help other implementers of this package not something that users should be aware of.

It's a `lib/src` library so end users won't see the doc anyway. Not exactly the same, but we do encourage using doc comments for information intended for maintainers in private APIs.
https://dart.dev/effective-dart/documentation#consider-writing-doc-comments-for-private-apis

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: I changed the name and annotated it with `@visibleForTesting`

pkgs/google_cloud_auth/test/test_utils.dart
brianquinlan: Empirically this seems to work... I thought that every `dart:io` method just throws on the web. That seems to be home the code is written: https://github.com/dart-lang/sdk/blob/da36aaa118b3dac2a065d4e6ec4782900a909938/sdk/lib/libraries.yaml#L479

Totally possible that I am holding this wrong.

kevmoo: This can fail on dart2wasm, though. the JS compilers are more lenient!

kevmoo: Double check...

natebosch: > this seems to work...

It's unlikely to work everywhere we want it to. I'm not sure we'd use this internally, but google3 disallows this style of usage.

There's been a lot more discussion outside this issue, but we did have intentions at one point to restrict this in more places (and now that I'm thinking of it, maybe we should do this for Dart 4)
https://github.com/dart-lang/sdk/issues/46917

I'd feel better if we handle the conditional imports here. We should be able to get the same effect by defining the non-io import to have this behavior.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: Done.

pkgs/google_cloud_auth/test/test_utils.dart
brianquinlan: Fixed with conditional imports.

pkgs/google_cloud_auth/lib/src/application_default_credentials.dart
brianquinlan: Done and I changed the other files in the package to match this style.

natebosch (approved):

PR #348, opened 2026-09-14
test(pubsub): run the Pub/Sub end-to-end tests against Google Cloud

sigurdm:
The eight Pub/Sub end-to-end tests were only ever run against the
emulator. They were already written to work against either backend
(`test/test_utils.dart` picks up application default credentials when
`PUBSUB_EMULATOR_HOST` is unset, and a couple of them already branch on
`isEmulator`), but nothing selected them in the Cloud Build integration
job, and the project did not have the API enabled in Terraform.

## Terraform

* Enable `pubsub.googleapis.com` in `.gcb/builds/services/main.tf`.

## Tests

* Tag the eight end-to-end tests `google-cloud` in addition to
  `firebase-emulator`. `.gcb/scripts/integration.sh` runs
  `dart test . -P google-cloud` from the repo root, so the root
  `dart_test.yaml` governs, and it already defines the tag and the
  preset. Nothing in `.gcb/` needs to change.

  I verified that a file tagged with both `firebase-emulator` and
  `google-cloud` really is selected by `-P google-cloud` from the repo
  root and by `-P firebase-emulator` from inside the package, and is
  still excluded from a plain `dart test`.

* Declare the `google-cloud` tag and preset in the package's own
  `dart_test.yaml` and exclude it by default. Without this, a plain
  `dart test` from inside `pkgs/google_cloud_pubsub` would run the tests
  against whatever project the developer's credentials point at.

* Replace the `DateTime.now().millisecondsSinceEpoch` name suffixes with
  a random suffix (`testResourceName`), following the pattern in
  `google_cloud_storage`'s `test_utils.dart`. Against the shared
  `dart-sdk-testing` project two builds can now overlap, and a shared
  name would make one of them fail with a `ConflictException` or observe
  the other's messages.

## Verification

`dart format`, `dart analyze` clean; `dart test .` and
`dart test -P firebase-emulator` (14 tests) pass locally.

TAG=agy


.gcb/builds/grants/main.tf
```diff
@@ -49,6 +49,15 @@ resource "google_project_iam_member" "sa-can-use-cloudtasks" {
   member  = "serviceAccount:${data.google_service_account.integration-test-runner.email}"
 }
 
+# The service account needs to create and delete topics and subscriptions, and
```
brianquinlan, line 52: You might need to change the

brianquinlan: This is not needed because it is controlled by terraform and we already have this enabled.

brianquinlan (approved):

.gcb/builds/grants/main.tf
sigurdm: Removed — thanks!

PR #352, opened 2026-09-18
feat(google_cloud_rpc): extract shared RetryRunner and ExponentialRetry

### LINKED ISSUE #346: Share retry logic between google_cloud_pubsub and google_cloud_storage (before pubsub 1.0)

sigurdm:
Extracts `RetryRunner`, `ExponentialRetry` (adding optional `jitter` and an injectable `isRetryable` predicate), `delaySequence` (computing `noRetriesAfter` eagerly when `delaySequence()` is called), `defaultIsRetryable`, `NoDelayRetry`, and `ChecksumValidationException` from `package:google_cloud_storage` into `package:google_cloud_rpc/retry.dart`.

- `package:google_cloud_storage` re-exports `package:google_cloud_rpc/retry.dart` for backwards compatibility.
- Moves `pkgs/google_cloud_storage/test/retry_test.dart` to `generated/google_cloud_rpc/test/retry_test.dart`.

Split out from #292 (1/4).
Closes #346

TAG=agy


kevmoo:
Filed #359 as a follow-up to wire `RetryRunner` and per-call `RequestOptions` into `ServiceClient._makeRequest()` and `sidekick/dart` once this lands.

